PR #18, opened 2016-10-11
Improve build

ashfire908:
Updated build documentation for both VS and Mono, as the instructions didn't work as-is. Also added the NUnit test adapter so tests can be run directly from visual studio.


marrony:
Were you not able to run tests through Visual Studio? I did all the job running directly from Visual Studio. I'm reinstalling VS right now since the "preview" period has expired and I will test again.

ashfire908:
When I ran it first, IIRC it didn't find any tests to run. I will retry as well.

marrony:
If it don't work again, try to look for NUnit addon on VS. Perhaps I've installed it and didn't bothered with it again.

ashfire908:
Ok, for clarity, VS does not run the NUnit tests without the NUnit test adapter installed, which I added via the project file for the tests project. The other changes I made were due to the instructions for building/testing via the command line being wrong, which is just a documentation update.

PR #19, opened 2016-10-11
Add coverage to CI

ashfire908:
WIP attempt to get coverage running.


ashfire908:
Opencover is documentation as not supporting being run under mono.

PR #25, opened 2017-02-01
Adds support for `@bytes` type

marrony:
close https://github.com/faunadb/sales-engineering/issues/508

### DIFF
--- a/FaunaDB.Client/Types/BytesV.cs
+++ b/FaunaDB.Client/Types/BytesV.cs
@@ -1,0 +1,83 @@
+﻿using System;
+using System.Linq;
+using System.Text;
+using FaunaDB.Query;
+using Newtonsoft.Json;
+
+namespace FaunaDB.Types
+{
+    /// <summary>
+    /// A FaunaDB bytes type.
+    /// <para>
+    /// See <see href="https://fauna.com/documentation/queries#values-special_types">FaunaDB Special Types</see>.
+    /// </para>
+    /// </summary>
+    public class BytesV : Value
+    {
+        byte[] _Value;
+
+        public byte[] Value
+        {
+            get
+            {
+                return (byte[])_Value.Clone();
+            }
+        }
+
+        internal BytesV(string base64)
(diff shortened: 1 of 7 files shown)


FaunaDB.Client/Types/BytesV.cs
```diff
@@ -0,0 +1,83 @@
...
+                return (byte[])_Value.Clone();
+            }
+        }
+
+        internal BytesV(string base64)
+        {
+            _Value = FromUrlSafeBase64(base64);
+        }
+
+        public BytesV(params byte[] value)
```
erickpintor: Does this constructor allow for an array of bytes to come in instead of a varargs?

FaunaDB.Client/Types/BytesV.cs
```diff
@@ -0,0 +1,83 @@
...
+            }
+
+            buffer.Append(")");
+
+            return buffer.ToString();
+        }
+
+        static byte[] FromUrlSafeBase64(string urlSafe)
+        {
+            var base64 = urlSafe.Replace('_', '/').Replace('-', '+');
```
erickpintor: Is this a standard transformation? Is there a changes where actual underscores or dashes shouldn't be replaced?

Perhaps we should put a comment explaining this.

marrony: This is not standard but it's a common transformation made by most of libraries.

https://docs.python.org/2/library/base64.html#base64.urlsafe_b64encode
https://en.wikipedia.org/wiki/Base64#URL_applications
http://stackoverflow.com/a/26354677

FaunaDB.Client/Types/BytesV.cs
marrony: yes

FaunaDB.Client/Types/BytesV.cs
erickpintor: Would you mind documenting that in the code for posterity?

PR #30, opened 2017-02-07
removes EqualityComparer from Result class

marrony:
close https://github.com/fauna/sales-engineering/issues/538

This PR might be a little controversial but when I was implementing support to @bytes type I realized that every time we had to add new data type we'd have to implement a comparer to support this type.

I did this for list/dictionary and I started to think that C# developers are very aware that collections don't override equals and they solved this by implementing extension methods so why should we do this.

Also other languages like scala a sentence like this `Some(Array(1)) == Some(Array(1))` would be false because arrays don't override equals as well.

This PR propose a new way of comparing `Result`s values by using `Match()` method and extract the value of the result. Also it's still override equals/hashcode by delegating to the underling type so if it's comparable it'll working as expected otherwise the developer is aware that the type is not comparable.

### DIFF
--- a/FaunaDB.Client.Test/ClientTest.cs
+++ b/FaunaDB.Client.Test/ClientTest.cs
@@ -864,8 +864,6 @@ namespace Test
 
         [Test] public async Task TestDatabase()
         {
-            Assert.ThrowsAsync<BadRequest>(async() => await adminClient.Query(Database("nonexistent-db")));
-
             await adminClient.Query(CreateDatabase(Obj("name", "database_for_database_test")));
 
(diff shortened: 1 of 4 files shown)


erickpintor:
What is the community standards to compare collections? Would you have some material for us to read?

marrony:
They use a extension methods https://msdn.microsoft.com/en-us/library/bb348567(v=vs.110).aspx to do it.

Unfortunately I could not find any article (even in Microsoft's page) explaining why they didn't implemented Equals/HashCode in collections, but looking at the [List](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Collections/Generic/List.cs) source code there's no doubt

PR #31, opened 2017-02-11
Fix parser + cleanups

marrony:
That PR fix a issue found in Json parser when an empty object literal was passed in like `{"@obj": {}}`. It also include some minor internal refactorings to make the types more consistent.


FaunaDB.Client/Collections/ImmutableDictionary.cs
```diff
@@ -1,81 +1,176 @@
...
+using System.Collections.Specialized;
 
 namespace FaunaDB.Collections
 {
+    class ImmutableDictionary<TKey, TValue> : IReadOnlyDictionary<TKey, TValue>
+    {
+        public static readonly ImmutableDictionary<TKey, TValue> Empty =
+            new ImmutableDictionary<TKey, TValue>();
+
+        readonly OrderedDictionary dictionary = new OrderedDictionary();
```
erickpintor: Did `OrderedDictionary` already existed in the standard library?

FaunaDB.Client/Types/ObjectV.cs
```diff
@@ -14,13 +14,14 @@ namespace FaunaDB.Types
...
+        public static readonly ObjectV Empty =
+            new ObjectV(ImmutableDictionary<string, Value>.Empty);
 
         public IReadOnlyDictionary<string, Value> Value { get; }
 
-        public ObjectV() : this(Empty.Value) { }
+        internal ObjectV() : this(Empty.Value) { }
 
-        public ObjectV(IReadOnlyDictionary<string, Value> value)
+        internal ObjectV(IReadOnlyDictionary<string, Value> value)
```
erickpintor: Did we drop all public constructors for `ObjectV`?

FaunaDB.Client/Collections/ImmutableDictionary.cs
marrony: Yes, but it's not generic

FaunaDB.Client/Types/ObjectV.cs
marrony: Yes, `ObjectV` instances can only be created through `Obj()` method

marrony: Ops, `ObjectV` is only created when parsing the response, `Obj()` create unwrapped objects

PR #34, opened 2017-03-11
Adds recursive refs

marrony:
Along with recursive refs, this PR introduce a new way of create sub clients with `NewWithOptions()` method. This method receives a variable list of options that enable the developer override for example the secret or enable/disable recursive refs.

Example:

```csharp
   var rootClient = new FaunaClient(secret: "secret");

   var childClient = rootClient.NewWithOptions(
      Options.Secret("new-secret"),
      Options.RecursiveRefs(true)
   );
```

ps: I still need to create another PR to introduce more api's enhancements (more cosmetics changes)


FaunaDB.Client/Client/DefaultClientIO.cs
```diff
@@ -18,12 +18,14 @@ namespace FaunaDB.Client
     class DefaultClientIO : IClientIO
     {
         readonly HttpClient client;
-        readonly string authHeader;
+        readonly AuthenticationHeaderValue authHeader;
+        readonly bool useRecursiveRefs;
 
-        internal DefaultClientIO(HttpClient client, string secret)
+        internal DefaultClientIO(HttpClient client, AuthenticationHeaderValue authHeader, bool useRecursiveRefs)
```
erickpintor: Using a flag like `useRecursiveRefs` to enabled api 2.1 can be a little bit dangerous as that api version could enable more than just the recursive refs as we extend its capabilities while still supporting the version 2.0.

It will be more clear if we treat this as a api version parameter, possibly a enum.

FaunaDB.Client/Client/DefaultClientIO.cs
```diff
@@ -34,11 +36,14 @@ public DefaultClientIO(Uri domain, TimeSpan timeout, string secret)
...
 
-            authHeader = AuthString(secret);
+            authHeader = AuthHeader(secret);
         }
 
         public IClientIO NewSessionClient(string secret) =>
-            new DefaultClientIO(client, secret);
+            new DefaultClientIO(client, AuthHeader(secret), useRecursiveRefs);
+
+        public IClientIO NewWithRecursiveRefs(bool useRecursive) =>
```
erickpintor: I think it is unlikely that people will bounce back and forth different versions of the db api during the application execution. That could also be risky in terms of data integrity, for example, if we require a data migration from one version to another.

Having methods that can induce the user to turn on and off a api version is a bad patter, IMO.

FaunaDB.Client/Client/FaunaClient.cs
```diff
@@ -59,7 +82,14 @@ internal FaunaClient(IClientIO root)
...
         /// <param name="secret">Auth token for the FaunaDB server.</param>
         public FaunaClient NewSessionClient(string secret) =>
-            new FaunaClient(clientIO.NewSessionClient(secret));
+            NewWithOptions(Options.Secret(secret));
+
+        /// <summary>
+        /// Creates a new <see cref="FaunaClient"/> with the options specified.
+        /// </summary>
+        /// <param name="options">Option modifiers to apply to the new client.</param>
+        public FaunaClient NewWithOptions(params Option[] options) =>
```
erickpintor: I think the same comment as `NewWithRecursiveRefs` applies here.

FaunaDB.Client/Types/RefParser.cs
```diff
@@ -0,0 +1,76 @@
+﻿namespace FaunaDB.Types
+{
+    public static class RefParser
+    {
+        public static RefV Parse(Value value)
+        {
+            return value.GetOption(Field.At("@ref")).Match(
```
erickpintor: I think if you know the api version ahead of time when creating the client you can break this down into two parsers: one for version 2.0, and one for version 2.1.

The parsing then becomes more assertive than attempted.

FaunaDB.Client/Types/RefV.cs
```diff
@@ -0,0 +1,157 @@
...
+        {
+            if (value == null)
+                throw new ArgumentNullException(nameof(value));
+        }
+
+        protected internal override void WriteJson(JsonWriter writer) =>
+            writer.WriteObject("@ref", Value);
+    }
+
+    public abstract class ObjectRef : RefV
```
erickpintor: I'm over the fence with this new abstractions. I would like to hear more about your motivation to create them.

He is my opinion:

As we create more classes to represent references, we expand the  required vocabulary the developer needs to know in order to use Fauna. It also give them the impression that classes, databases, keys are all different types of references instead of composed structured references. It increases complexity as now the driver must handle different types in order to express things that are actually the same thing.

Purity of abstraction would by us some simplicity, IMO. In the other hand, we also need some syntax sugar.

I was expecting something more on the lines of:

```csharp
new RefV(
  class: new RefV(BuiltInClasses.CLASSES, "my-class"),
  id: "123456789"
)
```

Perhaps a little more sugar on top of that constructor:

```charp
RefV.Of(
  class: RefV.Class("my-class"),
  id: "123456789"
)
```

Please, let me know your thoughts.

FaunaDB.Client/Types/RefParser.cs
marrony: That parser only parse 2.1 refs. The fact that it try to get a `@ref` first is because the nature of the new representation.

FaunaDB.Client/Types/RefV.cs
marrony: That abstraction was copied from core. The idea to copy the abstraction and use strong types instead of factory methods was to avoid mistakes that this model could cause. There are only a few combination of `id`, `class` and `database` attributes that are allowed by the api and also there's a distinction between class refs and database refs (if I understood correctly). I'm not against that model, to be honest my very first implementation was like that.

FaunaDB.Client/Client/DefaultClientIO.cs
marrony: I understand that is unlike people will bounce back and forth, but there must be a way to upgrade the version. All tests so far is based on 2.0 and it needs to keep working. What's your suggestion? Add the api version on constructor? And what about `NewSessionClient` method? Add the version there as well?

erickpintor: My suggestion would be to add the the API version to the constructor, probably as a constant, enum, or something like that. We should make 2.0 the default for now so the driver still works as it is and tests will pass.

That also makes it easy to setup the client as we make changes in the new api. We could set different parsers, for example.

The `NewSessionClient` should remain unchanged, IMO. That way, we make the user take the decision of which API version to use one place: the driver initialization only.

FaunaDB.Client/Types/RefParser.cs
erickpintor: I thought from the strategy document that all refs at version 2.1 would be objects, so you could just parse its properties. Did I miss anything?

FaunaDB.Client/Types/RefV.cs
erickpintor: I'm sure core have reasons to treat them as different things. Specially because they might have different behaviour internally. But at the surface, references are simpler to reason about with less boiler plate, IMO.

FaunaDB.Client/Types/RefParser.cs
marrony: No just the properties, one of the properties is the `@ref` as well.

```json
{
  "@ref": {
    "id": "self",
    "class": {
      "@ref": {
        "id": "a-class",
        "class": {"@ref": {"id": "classes"}}
      }
    }
  }
}
```

erickpintor: Ooow... recursiveness... 🤦‍♂️ 
Just ignore me.

FaunaDB.Client/Types/RefV.cs
freels: The point of the abstraction in core is to reduce the necessary exposure to the ref concept early on, hence the motivation for adding the query lang syntax sugar for creating schema objects, and schema refs. Internally everything is still a ref.

I think a good parallel is ruby's metaprogramming API vs its more traditional syntax for defining classes in source. Everything being an object is great, but it's a little mind-bending for daily use.

How that applies to the driver API, perhaps is still open, but they probably should mirror the query API functions as closely as possible.

erickpintor: So, I imagine that in C#, a call to the ref function to create structured refs would look something like:

```
Ref(Class("my-class"), "123")
```

If you need to manually create a structured ref value, I guess the most natural translation would be something like:

```
new RefV(new ClassV("my-class"), "123")
```

If that is the most natural path, I guess I would still ague about the class names as of now. `ObjectInstanceRef`, `CustomClassRef`, they sound a lot more scary than `new RefV(id: '123", class: new ClassV("some-class"))`.

FaunaDB.Client/Client/FaunaClient.cs
```diff
@@ -27,12 +27,28 @@ public class FaunaClient
...
+        /// <summary>
+        /// Constructs a new instance of <see cref="FaunaClient"/> with the specified arguments.
+        /// </summary>
         /// <param name="domain">Base URL for the FaunaDB server.</param>
         /// <param name="scheme">Scheme of the FaunaDB server. Should be "http" or "https". Default is "https"</param>
         /// <param name="port">Port of the FaunaDB server. Defaults to 443 if schema is "https" or 80 if schema is "http"</param>
         /// <param name="timeout">Timeout for I/O operations. Defaults to 1 minute.</param>
         /// <param name="secret">Auth token for the FaunaDB server.</param>
         /// <param name="clientIO">Optional <see cref="IClientIO"/>.</param>
+        [Obsolete("prefer FaunaClient(secret:endpoint:timeout:apiVersion:) instead")]
```
erickpintor: Is there a reason why we're replacing the driver's constructor?

FaunaDB.Client/Client/FaunaClient.cs
```diff
@@ -44,13 +60,15 @@ public class FaunaClient
...
-            this.clientIO = clientIO ??
-                new DefaultClientIO(new Uri($"{scheme}://{domain}:{port}"), timeout ?? TimeSpan.FromSeconds(60), secret);
+            this.clientIO = clientIO ?? CreateClient(secret, $"{scheme}://{domain}:{port}", timeout, null);
         }
 
-        internal FaunaClient(IClientIO root)
+        /// <summary>
+        /// Constructs a new instance of <see cref="FaunaClient"/> with the specified <see cref="IClientIO"/>.
+        /// </summary>
+        public FaunaClient(IClientIO clientIO)
```
erickpintor: Why are we making this public?

FaunaDB.Client/Types/Api21/RecursiveRef.cs
```diff
@@ -0,0 +1,123 @@
+﻿using System;
+using FaunaDB.Query;
+using Newtonsoft.Json;
+
+namespace FaunaDB.Types.Api21
+{
+    public abstract class ObjectRefV : RefV
```
erickpintor: Can we make this `internal` only?

FaunaDB.Client/Types/Api21/RecursiveRef.cs
```diff
@@ -0,0 +1,123 @@
...
+
+    public class ClassV : ObjectRefV
+    {
+        internal ClassV(string builtin) : base(builtin, null, null)
+        { }
+
+        public ClassV(string id, DatabaseV database = null) : base(id, BuiltIn.CLASSES, database)
+        { }
+
+        public static class BuiltIn
```
erickpintor: What do you think about pulling this out of `ClassV`?

FaunaDB.Client/Types/Api21/RecursiveRef.cs
```diff
@@ -0,0 +1,123 @@
...
+        { }
+    }
+
+    public class SelfV : ObjectRefV
+    {
+        public SelfV(ClassV @class) : base("self", @class, null)
+        { }
+    }
+
+    public class InstanceV : ObjectRefV
```
erickpintor: Is this the same as `Ref`? It seems like we're introducing a new term here.

FaunaDB.Client/Client/FaunaClient.cs
marrony: The reason is because in my option this constructor was bad designed. The last argument is exclusive, I mean, if you pass in a `ClientIO` it will ignore the others. That's way I also didn't retrofit it with api version.

FaunaDB.Client/Types/Api21/RecursiveRef.cs
marrony: no, because the other classes the derive from this is public

FaunaDB.Client/Types/Api21/RecursiveRef.cs
marrony: no, it serialize differently

FaunaDB.Client/Types/Api21/RecursiveRef.cs
marrony: The idea was to create some sort of dsl like `ClassV.BuiltIn.CLASSES`.

I don't know if it will be used by the users, theoretically the users could use it in `SelfV` and `InstanceV` but doesn't make sense to use with other types since we have specific types like `KeyV` and `FunctionV`

FaunaDB.Client/Client/FaunaClient.cs
marrony: for the reason explained above

FaunaDB.Client/Client/FaunaClient.cs
erickpintor: Ok. Make sure to add that information to the change log for the next version, so we don't forget.

FaunaDB.Client/Types/Api21/RecursiveRef.cs
erickpintor: Is that a language constraint?

FaunaDB.Client/Types/Api21/RecursiveRef.cs
erickpintor: I guess I would just prefer to keep the types simpler. BuiltIn could be it's own thing.

FaunaDB.Client/Types/Api21/RecursiveRef.cs
erickpintor: I guess my point is that once we turn 2.1 the default version, this will be the only valid form of a reference. Is it better if people still refer to this as a ref? I'm worried about adding a new term on the drivers that doesn't match with the wire protocol.

FaunaDB.Client/Types/Api21/RecursiveRef.cs
marrony: yes, the base class cannot be less accessible than its derived

`Inconsistent accessibility: base class FaunaDB.Types.Api21.ObjectRefV is less accessible than class FaunaDB.Types.Api21.DatabaseV`

FaunaDB.Client/Types/Api21/RecursiveRef.cs
marrony: I understand that once we enable 2.1 it will be the default ref, but this is still optional. The old ref need to keep working and also the new and old have different shapes: `{"@ref": "indexes/all_widgets"}` => `{"@ref": {"id": "all_widgets", "class": {"@ref": "classes"}}}`

erickpintor: Do we support refs 2.0 on version 2.1 of the API?

erickpintor: I guess this is regardless of supporting old syntax or not.

What I'm thinking is that people will be using references and out of the sudden they will have to refer to the same things as being "instances". It's a mind shift that I want to make sure we think through it rather than accidentally making that shift because of a implementation choice.

To clarify our options: Is it possible to use the name `RefV` in different namespaces? How does C# handle name collision?

Would that make the transition between API versions easier considering people won't need to adopt another terminology?

freels: We do support them but there's no need to support both in a driver that is using the 2.1 api

marrony: > Is it possible to use the name RefV in different namespaces? How does C# handle name collision?

yes, it's possible

marrony: But it will make the life of developers harder, because whenever you have the "old" namespace imported, the developer would have to use FQN for the new objects.

```csharp
RefV _ref = new FaunaDB.Types.Api21.RefV("12345");
```

erickpintor: Is there any precedence or it always need the FQN when names colide?

FaunaDB.Client/Client/DefaultClientIO.cs
```diff
@@ -105,5 +98,17 @@ static string QueryString(IReadOnlyDictionary<string, string> query)
...
         }
+
+        static HttpClient CreateClient(Uri endpoint, TimeSpan timeout)
+        {
+            var client = new HttpClient();
+            client.BaseAddress = endpoint;
+            client.Timeout = timeout;
+            client.DefaultRequestHeaders.AcceptEncoding.Add(new StringWithQualityHeaderValue("gzip"));
+            client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+            client.DefaultRequestHeaders.Add("X-FaunaDB-API-Version", "2.1");
```
erickpintor: Just as a note: I think we should create a branch of master like `release-1.0` and add the same line pinning the api version to 2.0.

FaunaDB.Client/Types/RefParser.cs
```diff
@@ -0,0 +1,70 @@
...
+        static RefV Mk(IOption<Value> id, IOption<Value> cls, IOption<Value> db)
+        {
+            var idE = ParseId(id);
+            var classE = Cast<ClassV>(cls);
+            var databaseE = Cast<DatabaseV>(db);
+
+            if (classE == null && databaseE == null)
+                return BuiltIn.FromName(idE);
+
+            if (classE == null)
```
erickpintor: This seems a little odd to me. What are the casses where `classE` could be null?

I was expecting a check for each `BuiltIn` and a `new RefV(...)` at the end if none applies...

FaunaDB.Client/Types/RefParser.cs
```diff
@@ -0,0 +1,70 @@
...
+            if (BuiltIn.INDEXES == classE)
+                return new IndexV(id: idE, database: databaseE);
+
+            if (BuiltIn.FUNCTIONS == classE)
+                return new FunctionV(id: idE, database: databaseE);
+
+            if (BuiltIn.KEYS == classE)
+                return new KeyV(id: idE, database: databaseE);
+
+            if (databaseE == null)
```
erickpintor: Same question as above.

FaunaDB.Client/Types/RefParser.cs
```diff
@@ -0,0 +1,70 @@
...
+                return new RefV(id: idE, @class: classE);
+
+            return null;
+        }
+
+        static string ParseId(IOption<Value> id)
+        {
+            return id.FlatMap(value =>
+            {
+                return value.To(Codec.STRING).Match(
```
erickpintor: Is this check still required for 2.1? I was under the impression that ref's ids would always be strings from now on.

FaunaDB.Client.Test/DeserializationTest.cs
```diff
@@ -115,5 +114,126 @@ static void AssertJsonEqual(Value value, string json)
...
+            AssertJsonEqual(
+                new RefV(
+                    id: "123456789",
+                    @class: new ClassV(
+                        id: "child-class",
+                        database: new DatabaseV("child-database"))),
+                "{\"@ref\": {\"id\": \"123456789\", \"class\": {\"@ref\": {\"id\": \"child-class\", \"class\": {\"@ref\": {\"id\": \"classes\"}}, \"database\": {\"@ref\": {\"id\": \"child-database\", \"class\": {\"@ref\": {\"id\": \"databases\"}}}}}}}}"
+            );
+
+            // Test id with long value
```
marrony: The id is of different types in json

```json
{"@ref": {"id": "123456789"}}
```

```json
{"@ref": {"id": 123456789}}
```

FaunaDB.Client/Types/RefParser.cs
marrony: According to our documentation it could be an string or a number

FaunaDB.Client/Types/RefParser.cs
marrony: you're right

CHANGELOG.txt
```diff
@@ -1,3 +1,6 @@
+2.0.0-SNAPTHOT
+* Deprecate FaunaClient(secret:domain:scheme:port:timeout:clientIO:) constructor
```
erickpintor: We didn't deprecate that constructor. It was removed.
We also need information about the new data structure, the removal of Ref(string) syntax sugar, the API version supported, etc.

FaunaDB.Client/Types/RefParser.cs
erickpintor: You're right. It could be "classes", for example.

CHANGELOG.md
```diff
@@ -0,0 +1,35 @@
+## 2.0.0-SNAPSHOT
+
+- Adds support for new reference types in FaunaDB API 2.1
+  - `DatabaseV`, `ClassV`, `IndexV`, `FunctionV` and `KeyV`
+- Removes `Ref(string id)` function
```
erickpintor: We need to say that references will have to be created with `new RefV` from now on.

FaunaDB.Client/Types/RefV.cs
```diff
@@ -0,0 +1,159 @@
...
+            return $"{GetType().Name}(id = \"{Id}\"{cls}{db})";
+        }
+    }
+
+    static class RefExtensions
+    {
+        static bool IsBuiltIn(this RefV _ref) =>
+            _ref.Class == null && _ref.Database == null;
+
+        public static bool IsClassRef(this RefV cls)
```
erickpintor: Can we make this internal? Otherwise I would remove it and let the server validate it.

FaunaDB.Client/Types/RefV.cs
```diff
@@ -0,0 +1,159 @@
...
+        {
+            if (IsBuiltIn(cls))
+                return true;
+
+            return cls.Class != null && IsClassRef(cls.Class);
+        }
+
+        public static bool IsDatabaseRef(this RefV db)
```
erickpintor: Same consideration as above.

FaunaDB.Client/Types/RefV.cs
```diff
@@ -0,0 +1,159 @@
...
+        public DatabaseV(string id, DatabaseV database = null) : base(id, BuiltIn.DATABASES, database)
+        { }
+    }
+
+    /// <summary>
+    /// A Class reference. Terse version for <c>new RefV(id, BuiltIn.CLASSES, database)</c>
+    /// </summary>
+    public class ClassV : RefV
+    {
+        internal ClassV(string builtin) : base(builtin, null, null)
```
erickpintor: Can we move this to `RefV`? It would be more consistent since we're expressing builtin references.

FaunaDB.Client/Types/RefParser.cs
erickpintor: I was digging into this. Core will always return ref's id as a `String`. We can remove this check.

FaunaDB.Client/Types/RefV.cs
marrony: the method is public otherwise classes cannot access it, but the class is private

FaunaDB.Client/Types/RefV.cs
marrony: if we move we wouldn't be allowed to instantiate `ClassV` using this constructor

marrony: That was a special constructor I added to make all builtin classes of type `ClassV`, it's not necessary though, we can instantiate `RefV` types

FaunaDB.Client/Types/RefV.cs
erickpintor: So, people won't see these methods. Correct?

FaunaDB.Client/Types/RefV.cs
erickpintor: I would remove it or move it to `RefV`, if needed. Having it here is a little confusing, even as a internal constructor.

FaunaDB.Client/Types/RefV.cs
marrony: correct

FaunaDB.Client/Types/RefParser.cs
```diff
@@ -0,0 +1,50 @@
...
+            if (BuiltIn.INDEXES == classE)
+                return new IndexV(id: idE, database: databaseE);
+
+            if (BuiltIn.FUNCTIONS == classE)
+                return new FunctionV(id: idE, database: databaseE);
+
+            if (BuiltIn.KEYS == classE)
+                return new KeyV(id: idE, database: databaseE);
+
+            return new RefV(id: idE, @class: classE, database: databaseE);
```
marrony: We can also stop using sugar classes here and just instantiate a `RefV`, the above lines would be removed and let only user use those classes. Thoughts?

erickpintor: I like the type information. Just so people can do a `instanceOf` if they need to.

FaunaDB.Client.Test/ErrorsTest.cs
```diff
@@ -36,16 +36,16 @@ namespace Test
...
 
         [Test] public async Task TestPermissionDenied()
         {
-            var key = await rootClient.Query(Create(Ref("keys"), Obj("database", DbRef, "role", "client")));
+            var key = await rootClient.Query(CreateKey(Obj("database", DbRef, "role", "client")));
 
             var client = GetClient(secret: key.Get(Field.At("secret").To(Codec.STRING)));
 
-            AssertQueryException<PermissionDenied>(client, Paginate(Ref("databases")), "permission denied", "Insufficient privileges to perform the action.");
+            AssertQueryException<PermissionDenied>(client, Paginate(new RefV("databases")), "permission denied", "Insufficient privileges to perform the action.");
```
erickpintor: Can you use the BuiltIn here?

FaunaDB.Client/Types/RefV.cs
```diff
@@ -0,0 +1,156 @@
...
+            {
+                case "classes": return CLASSES;
+                case "indexes": return INDEXES;
+                case "databases": return DATABASES;
+                case "keys": return KEYS;
+                case "functions": return FUNCTIONS;
+                case "tokens": return TOKENS;
+            }
+
+            throw new NotSupportedException($"The builtin class '{name}' is not supported");
```
erickpintor: We should not fail if the server returns a different builtin.

FaunaDB.Client/Types/RefV.cs
```diff
@@ -0,0 +1,156 @@
...
+        public RefV Class { get; }
+        public RefV Database { get; }
+
+        public RefV(string id, RefV @class = null, RefV database = null)
+        {
+            Id = id;
+            Class = @class;
+            Database = database;
+
+            if (@class != null && !@class.IsClassRef())
```
erickpintor: I understand that the developer could build a wrong ref without this validations but, I think this should be the db's responsibility.

CHANGELOG.md
```diff
@@ -0,0 +1,36 @@
...
+  - All refs from now on should be created using constructors like `new RefV(...)` or `new DatabaseV(...)`
```
marrony: you mean `ref:id:` function?

erickpintor: Yes. That one.

marrony: Yeah, it's still available. Will just add more test on top of the function. I'm just testing serialization

marrony:
In this PR I included a refactoring on json parser, I'll remove that refactoring and open a separate PR bc there are some changes I'll need to include on version 1.0 as well.

marrony:
Does anyone would like to take another look in this PR?

erickpintor:
I don't think we got a second review yet. @eaceaser would be a good candidate.

FaunaDB.Client/Types/RefV.cs
```diff
@@ -0,0 +1,126 @@
...
+    /// A FaunaDB ref type.
+    /// <para>
+    /// See <see href="https://fauna.com/documentation/queries#values-special_types">FaunaDB Special Types</see>.
+    /// </para>
+    /// </summary>
+    public class RefV : Value
+    {
+        public string Id { get; }
+        public RefV Class { get; }
+        public RefV Database { get; }
```
eaceaser: this may be my ignorance of the recursive ref spec, but is `Class` and `Database` here always a `ClassV` and a `DatabaseV`? Or will they be arbitrary RefVs?

marrony: yes and no, it might be an instance of `RefV` with the same shape of `ClassV`, given `ClassV` is just a sugar, but yes, we can constraint ourselves be a `ClassV` with no harm. I've implemented using `ClassV` and `DatabaseV` in the beginning but for some reason I don't remember now I changed.

erickpintor: In theory those are `ClassV` and `DatabaseV` but in reality, those are any reference that the server returns. I doesn't make sense to restrict the type, IMO.

eaceaser:
This looks good to me! Feel free to merge it at  your leisure

PR #37, opened 2017-03-27
Serialize/Deserialize user classes

marrony:
This is a response for a "resquest" from a user in gitter that asked for the ability to serialize its own classes instead of having to convert manually. In this PR I added the ability to not only serialize but to deserialize as well.

```csharp
class Person
{
   public string Name;
   public LongV Age;
   ....
}

// serialize
await client.Query(Obj(new Person("John", 30)));

//deserialize
Person person = await client.Query<Person>(Obj("Name", "John", "Age", 30));

List<Person> people = await client.Query<List<Person>>(Arr(
   Obj("Name", "John", "Age", 30), Obj("Name", "Mary", "Age", 30)
));
```


erickpintor:
I was expecting this to be part of the `Codec` interface rather than the `Client` interface. Something like: `await client.Query(...).To(Person)`

Question, can this handle composed instances? Like: `class Person { public Address address }`

marrony:
Put this on Codec would involve to much introspection and object manipulation, thing the json framework do for free, and yes this can handle pretty much any object composition.

erickpintor:
Can you traverse the object tree and than decode the underneath value into a custom class?

For example:

```csharp
var res = await client.Query(...)
Person firstFriend = res.Get("person").Get("fiends", 0) // PS: I don't quite remember the API here.
```

freels:
Does this correctly handle our special types such as Refs?

marrony:
@erickpintor No, currently our api doesn't support this, for this we would have to implement a codec anyway.

@freels Yes, we can mix-and-match primitive types, user defined types and our types.

freels:
In other drivers we're careful not to leak the fact that the wire protocol is JSON, so we should to be sure to do that here as well. (I think scala is a good example here).

marrony:
We're not leaking this info explicitly, however we're using a json framework under the hood anyway, and this framework have this feature, that we're already using when parsing our internal types. What I did was leverage this feature exposing a extra generic `Query<T>()` method

freels:
Yeah, that makes sense, however, I suspect that in order to exert control over the serialization format the user is going to have to use the json libraries annotations api, right? Or do they have no control over that?

Also, IMO it shouldn't be possible to decode the structure of a ref by reaching into its "@ref" field, for example.

marrony:
That's why I say "explicitly", If user would like to control this they would have to use the framework annotations. And yes, if the user annotate some field with "@ref" they would take control over this inadvertently (or not).

I'm trying to find some library to help introspect the classes/objects without having to use the low-level api of c# but I'm not having much success. So I think we would have to create a very basic implementation and force user to have a default constructor on its classes and only access public getter/setters and perhaps an annotation to override the field name. I'll take a look on Go driver to have more ideas. Thoughts?

jchris:
Can we do a code search and see if namespace conflicts with our reserved names in annotation land are common in practice?

eaceaser:
@freels the scala driver doesn't do arbitrary user-defined-type serialization/deserialization though, does it? without defining your own decoder/encoder?

We're kind of in the same boat in Scala then, right? Even though Scala is nice in the sense that we can create a codec for `Product` and have it apply to case classes, to solve the general case we have to define all our own interface, annotations etc., even if we do just delegate to Jackson under the hood.

freels:
@eaceaser The scala driver sorta follows what you get from scala-json, so yes, you have to define your own codec typeclass instance. IME this is expected in scala land (as opposed to an annotation-based API). The faunadb-scala codecs are composeable as well, so the user can define their own custom type codec heirarchy.

freels:
The main thing is that the scala and java APIs do not expose a generic JSON data AST, but rather a fauna data AST. This helps prevent edge cases such as decoding parts of the special type internals and ensure that round-tripping doesn't break when our non-json types are involved.

eaceaser:
Yeah I agree with you about the path forward. I guess what I'm trying to figure out is: I'm not 100% sure if reimplementing the whole scala codec chain in C# is a good idea (or if its even possible to do effectively) - I've been out of the C# game for too long.

I know the C# world is more used to attribute-based processing right? I guess my guidance is that we should do the harder work now, and implement our own attributes for serialization purposes. These can proxy to Json.NET attributes, but we should define our own interface to abstract over the underlying serialization format.

Thats my 2 cents on the matter

freels:
Ok, cool. I think you're right that some sort of annotation-based API is more idiomatic, but I don't really know.

marrony:
@eaceaser  today we have a codec schema that can be composed as well, we would just need to create some composition helpers like `Field` and `Map` and we would be able to do things like:

```csharp
var personCodec = Map(
   (string name, long age) => new Person(name, age),
   Field("name", Codec.STRING),
   Field("age", Codec.LONG)
);

var person = ObjectV.With("name", "John", "age", 42).To(personCodec);
```

erickpintor:
The important thing right now is to find the most idiomatic form of custom data structures (de)serialization in each environment.

Java folks expect annotations, Scala folks expect type classes(?), Go relies heavily on reflection. What about C#? Seems like attribute based processors? (I don't really know).

I agree with @eaceaser that we should take the time and implement this right. IMO, we should offer our own semantics no matter what we use behind the scenes to implement it.

I also think that this should be part of the transversal API rather than the client interface. The query language is powerful enough so you can get a lot from a single query. If people have to restrict themselves to a single `T` type two things can happen: people will fire more queries to fetch individual types; or there will be an expansion of DTOs polluting the code in order to be able to use the type safe query interface.

eaceaser:
👍 to Erick's mention of making sure the type paramaterization is at the response object traversal API.

A couple things about this:

* I think the way most people will be familiar with for object mapping will be similar between C# and Java:

1. Sane default behavior for mapping plain-old objects: i.e. mapping builtin types, nested objects, etc. properly.
2. Attributes/annotations to override the default behavior in simple ways: e.g. opt-in vs. opt-out, field renames, field ignores, etc.
3. The ability to set custom serializers/deserializers for more complex operations.

Downside here is that writing a full-featured mapper API like this is pretty intensive, with all the corner cases. For example, Json.NET, msgpack, etc. all implement their own versions of this, and its a fair amount of code. I'm concerned that going down this route will lead to a slippery slope where we essentially have to abstract around one of these underlying serializers, but maintain a large API surface area to cover a bunch of corner cases.

* With that being said, I don't necessarily think the Codec stuff we're doing in Java and C# is bad. The annoying part is there's a non-trivial amount of boilerplate required to serialize an object, but is that a huge downside? So maybe the first step is to do what @marrony suggests and extend the Codec API to allow for object serialization/deserialization?

marrony:
as I said I was looking for some library to help us on this task and I found [AutoMapper](https://www.nuget.org/packages/AutoMapper/), it might help (or not) us somehow

eaceaser:
I've looked at automapper before. It's API is very similar to our Codec stuff IIRC. Since our problem is relatively constrained (we don't need to support translating from one arbitrary object to another, just one arbitrary object to an Value type) I think we can roll our own thing

erickpintor:
Implementing helpers is a palliative solution but, people will still be looking for the optimal solution for each language's driver.

I would say we should tackle the right problem. I'm also OK if we want to use a library underneath, as long as at the surface, people use our own annotations and APIs.

eaceaser:
On second thought, I don't think we're going to be able to directly piggyback off of another serializer (at least, not easily). Any attributes that a user has for their own serialization will run the risk of conflicting with our own use of the serializer.

I think the right solution then is to map to our value types, then serialize those in a context that we control.

So, in other words, I think we should start with implementing default object mapping behavior. We can copy what things like Automapper does, but we should, as a first pass, be able to take an arbitrary plain-old object and transform it to our Value types without additional configuration. @erickpintor does that sound reasonable?

erickpintor:
I think it does. That's pretty much what we have done for Go driver. Their reflection API is complex to use but we're lucky that we don't have a lot of types to cover.

PR #38, opened 2017-03-30
Adds support for user classes serialization

marrony:
In this PR I just implemented the serialization of user classes, the deserialization I will do in a separate PR. In summary, the user can do now:

```csharp
class Spell
{
    [FaunaDBField("name")]
    public string Name { get; }

    [FaunaDBField("element")]
    public string Element { get; }

    [FaunaDBField("cost")]
    public int Cost { get; }

    public Spell(string name, string element, int cost)
    {
        Name = name;
        Element = element;
        Cost = cost;
    }
}

var spell = new Spell("Magic Missile", "arcane", 10);

var spellCreated = await client.Query(
    Create(
        Ref("classes/spells"),
        Obj("data", Encoder.EncodeObject(spell))
    )
);
```

I included two annotations, `FaunaDBField` and `FaunaDBIgnore`. I don't know if it's a good name or we can simplify to just `Field` and `Ignore`.

Along with the primitive types users can also use fauna types in their classes.

In theory they can use any type that implement the interfaces `IEnumerable` and `IDictionary`, so pretty much all collections types and primitive arrays are supported.


FaunaDB.Client/Attributes/FaunaDBAttributes.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+
+namespace FaunaDB.Attributes
+{
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
+    public class FaunaDBFieldAttribute : Attribute
```
erickpintor: I personally prefer `Field` and `Ignore`. Less boilerplate to type.

FaunaDB.Client/FaunaDB.Client.csproj
```diff
@@ -87,5 +87,7 @@
     <Compile Include="Query\PageHelper.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="Types\BytesV.cs" />
+    <Compile Include="Reflection\Encoder.cs" />
```
erickpintor: Not sure if I like `Reflection` namespace. We're not providing a reflection API.
`Codec`? `Encoding`? Idk.

FaunaDB.Client/Reflection/Encoder.cs
```diff
@@ -0,0 +1,172 @@
...
+using System.Reflection;
+using FaunaDB.Query;
+using FaunaDB.Types;
+using FaunaDB.Attributes;
+
+namespace FaunaDB.Reflection
+{
+    public static class Encoder
+    {
+        public static Expr EncodeObject(object obj) =>
```
erickpintor: `EncodeObject` gives the impression that I can't encode anything else, like lists for example. Is that correct?

FaunaDB.Client/Reflection/Encoder.cs
```diff
@@ -0,0 +1,172 @@
...
+        public Expr Encode(object obj)
+        {
+            if (obj == null)
+                return NullV.Instance;
+
+            if (typeof(Expr).IsInstanceOfType(obj))
+                return (Expr)obj;
+
+            if (stack.Contains(obj))
+                throw new InvalidOperationException($"Self referencing loop detected for object `{obj}`");
```
erickpintor: Nice!

FaunaDB.Client/Reflection/Encoder.cs
```diff
@@ -0,0 +1,172 @@
...
+                    if (typeof(IDictionary).IsInstanceOfType(obj))
+                        return WrapDictionary((IDictionary)obj);
+
+                    if (typeof(IEnumerable).IsInstanceOfType(obj))
+                        return WrapEnumerable((IEnumerable)obj);
+
+                    return WrapObj(obj);
+            }
+
+            return NullV.Instance;
```
erickpintor: Shouldn't this be a `default` at the `switch`?

erickpintor: Or maybe an exception since it's a non recognized type?

FaunaDB.Client/Reflection/Encoder.cs
marrony: Actually it's possible given I had to "encode" fields/properties of any type and that function is called recursively for each field/property if you pass in a collection of any type it'll encode as well

FaunaDB.Client/Reflection/Encoder.cs
marrony: an exception would make more sense, given I tested every possible enum values in the switch

FaunaDB.Client/FaunaDB.Client.csproj
marrony: Idk either, let ask for another dev to give more thoughts

FaunaDB.Client/Reflection/Encoder.cs
erickpintor: So, I guess we could call it just `Encode` then?

FaunaDB.Client/Reflection/Encoder.cs
erickpintor: Yeah. It makes sense to be an Exception.

marrony:
changed the namespace from `Reflection` => `Encoding` and the encoding method from `EncodeObject` => `Encode`

marrony:
The last commit I pushed I improved serialization performance by creating (and caching) a specialized lambda to convert a given type. Instead of introspect the type every time it needs to be serialized, it introspect only in the first time and generate a dynamic lambda that access the fields and properties as if the developer had created that function.

for example, for the given class:
```csharp
class User
{
   public string UserName { get; }
   public string Password { get; }
}
```
it will generate a lambda similar to this:
```csharp
Func<User, Expr> UserToExpr = user => {
   var dic = new Dictionary<string, Expr> {
      { "UserName", Encode(user.UserName) },
      { "Password", Encode(user.Password) }
   };
   return Wrap(dic);
};
```

a simple test shows a significant improvement
```csharp
class Foo
{
    public string foo = "foo";
}

class Bar
{
    public Foo foo = new Foo();
}

var bar = new Bar();
var begin = DateTime.Now;
for (int i = 0; i < 2000000; i++)
    Encode(bar);
var end = DateTime.Now;

var span = end - begin;
Console.WriteLine(span);
```

the above test, run 2 millions encoding of the `Bar` class
* Without dynamic lambda: `00:00:50.1056060`
* With dynamic lambda: `00:00:04.6676160`

FaunaDB.Client/Encoding/Encoder.cs
```diff
@@ -0,0 +1,244 @@
...
+using FaunaDB.Query;
+using FaunaDB.Types;
+using FaunaDB.Attributes;
+using System.Linq.Expressions;
+
+namespace FaunaDB.Encoding
+{
+    public static class Encoder
+    {
+        public static Expr Encode(object obj) =>
```
erickpintor: I think this should work at the `Value` level. I don't see a reason why automatic serialization should return `Expr`. Do you have any thoughts on that?

FaunaDB.Client/Encoding/Encoder.cs
```diff
@@ -0,0 +1,244 @@
...
+
+        public Expr Encode(object obj)
+        {
+            if (obj == null)
+                return NullV.Instance;
+
+            if (typeof(Expr).IsInstanceOfType(obj))
+                return (Expr)obj;
+
+            if (stack.Contains(obj))
```
erickpintor: How does `Contains` check for equality? Do we have the risk of haring object pools returning `true` for equal objects but at different pointers? Similar to comparing strings in java with `==`?

FaunaDB.Client/Encoding/Encoder.cs
```diff
@@ -0,0 +1,244 @@
...
+                    if (typeof(IDictionary).IsInstanceOfType(obj))
+                        return WrapDictionary((IDictionary)obj);
+
+                    if (typeof(IEnumerable).IsInstanceOfType(obj))
+                        return WrapEnumerable((IEnumerable)obj);
+
+                    return WrapObj(obj);
+            }
+
+            throw new InvalidOperationException($"Could not encode object `{obj}`");
```
erickpintor: Maybe print more information. Like: `Unsupported type {X} at {obj}`

FaunaDB.Client/Encoding/Encoder.cs
marrony: Under the hood it will use an `EqualityComparer` that will check if the type implements `IEqualiable` if not it will use `Equals()` and `GetHashCode()`, thinking better now it might be an issue because if we have have an object that implements `Equals()` and `GetHashCode()` and it happens to be true for different instances it will throw the exception. I'll use a overloaded version of `Contains` that receive an custom comparer to check only for references

erickpintor: Nice! Let me know the results once you're done.

marrony: I was able to simulate and created an test case to cover that case

FaunaDB.Client/Attributes/FaunaDBAttributes.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+
+namespace FaunaDB.Attributes
+{
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
+    public class FieldAttribute : Attribute
```
erickpintor: Missing docs?

FaunaDB.Client/Attributes/FaunaDBAttributes.cs
```diff
@@ -0,0 +1,19 @@
...
+        public string Name { get; }
+
+        public FieldAttribute(string name)
+        {
+            Name = name;
+        }
+    }
+
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
+    public class IgnoreAttribute : Attribute
```
erickpintor: Docs.

FaunaDB.Client/Encoding/Encoder.cs
```diff
@@ -0,0 +1,258 @@
...
+using System.Reflection;
+using FaunaDB.Types;
+using FaunaDB.Attributes;
+using System.Linq.Expressions;
+
+namespace FaunaDB.Encoding
+{
+    using Converter = Func<EncoderImpl, object, Value>;
+
+    public static class Encoder
```
erickpintor: This class need docs.

FaunaDB.Client/Encoding/Encoder.cs
```diff
@@ -0,0 +1,258 @@
...
+    {
+        public static Value Encode(object obj) =>
```
erickpintor: This method too. Specially explaining how it works.

FaunaDB.Client/Encoding/Encoder.cs
```diff
@@ -0,0 +1,258 @@
...
+
+            return member.Name;
+        }
+
+        class ReferenceComparer : IEqualityComparer<object>
+        {
+            public static readonly ReferenceComparer Default = new ReferenceComparer();
+
+            public new bool Equals(object x, object y) =>
+                object.ReferenceEquals(x, y);
```
erickpintor: Were you able to confirm if this will work correctly considering the String Intern mechanisms in C#?
I guess it works considering I've seem a test using the same string repeatedly.
I wonder if something like this would break the implementation:

```
var a = "a string"
var b = a

Encode(SomeClass(a, b))
```

FaunaDB.Client/FaunaDB.Client.csproj
```diff
@@ -87,5 +87,7 @@
     <Compile Include="Query\PageHelper.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="Types\BytesV.cs" />
+    <Compile Include="Encoding\Encoder.cs" />
```
erickpintor: I was thinking about this again. Maybe we should make this part of `Types` just so people doesn't have to import another namespace. I imagine this will be used way more than the alternative.

FaunaDB.Client/Encoding/Encoder.cs
marrony: That shouldn't be a problem specially for strings, because it's a primitive type so we don't recurse into the struct/class fields. Also even if we use the very same reference in two or more fields of a class again it won't be a problem because it will encode each field recursively and when it return the recursion it removes the reference from the stack.

FaunaDB.Client/FaunaDB.Client.csproj
marrony: And what about the attributes? Currently it's in `FaunaDB.Attributes` namespace so should user want to override a field name or ignore a field they would have to import that namespace anyway

erickpintor: I would move that also. So we keep the boilerplate to use fauna short.

marrony:
Just one final commit to change `Expr` types to `Value` types on tests

FaunaDB.Client/Types/ObjectV.cs
```diff
@@ -86,6 +86,9 @@ public override string ToString()
         public static ObjectV With() =>
             new ObjectV();
 
+        public static ObjectV With(IReadOnlyDictionary<string, Value> values) =>
```
marrony: Also I had to add this factory method to create arbitrary sized objects

FaunaDB.Client/Types/Encoder.cs
```diff
@@ -0,0 +1,278 @@
...
+            return new ObjectV(ret);
+        }
+
+        Value WrapObj(object obj)
+        {
+            var type = obj.GetType();
+
+            Converter converter;
+
+            lock (converters)
```
eaceaser: since this is Write-seldom, read-often, should we use a https://msdn.microsoft.com/en-us/library/system.threading.readerwriterlockslim(v=vs.110).aspx here?

marrony: let me check

marrony: Ok, in my test I ran 100 threads and each thread did 20000 encodings. Using `locks` it ran at `00:00:27.9190770` and using `ReaderWriterLockSlim` it ran at `00:00:26.4105080`.

My conclusion is that it won't make much difference using `lock` compared to `ReaderWriterLockSlim` given the contention is too low and the overhead added by `ReaderWriterLockSlim` is kinda high.

Please have look in my test and see if I'm doing something completely wrong

```csharp
        public void TestThread()
        {
            var bar = new Bar();

            int maxThreads = 100;

            ParameterizedThreadStart thread = delegate(object obj)
            {
                for (int i = 0; i < 2000000 / maxThreads; i++)
                    Encode(bar);
            };

            Thread[] threads = new Thread[maxThreads];
            for (int i = 0; i < maxThreads; i++)
                threads[i] = new Thread(thread);

            var begin = DateTime.Now;

            for (int i = 0; i < maxThreads; i++)
                threads[i].Start();

            for (int i = 0; i < maxThreads; i++)
                threads[i].Join();
            
            var end = DateTime.Now;

            var span = end - begin;
            Console.WriteLine(span);
        }
```

eaceaser: If you set maxThreads to the number of cores you have on your machine (or two times the number of cores), does it change at all? You may be dominated by context switching.

eaceaser: but i think its fine regardless, im just curious at this point

marrony: I reduced to maxThreads to 2 and the results was: 
* lock: `00:00:10.2585850`
* ReaderWriterLockSlim: `00:00:09.6747470`

eaceaser: word. seems like the RWLSlim is a little faster but its up to you if you wanna change it

eaceaser:
Feel free to merge this whenever, this looks good

PR #39, opened 2017-04-05
Adds support for user classes deserialization

marrony:
In that PR I included a new attribute `Creator` to annotate constructors or static method in order to instruct the decoder which one to use in order to instantiate user objects. I also included a property to specify default values on attribute `Field`


FaunaDB.Client/Types/Value.cs
```diff
@@ -81,7 +81,35 @@ public abstract class Value : Expr
...
+        /// Attempts to coerce this value to given type T specified.
+        /// <para>
+        /// See <see cref="Decoder"/>
+        /// </para>
+        /// </summary>
+        /// <exception cref="System.InvalidOperationException">
+        /// Thrown if this instance of <see cref="Value"/> is not convertible to T.
+        /// Or if type T does not fullfil the requirements to be decoded.
+        /// </exception>
+        public T To<T>() =>
```
erickpintor: Shouldn't this return a `IResult` to keep the api consistent.

FaunaDB.Client/Types/Decoder.cs
```diff
@@ -0,0 +1,469 @@
...
+            for (var i = 0; i < array.Count; i++)
+                ret.SetValue(Decode(array[i], elementType), i);
+
+            return ret;
+        }
+
+        static T FromScalar<T, R>(Value value)
+        {
+            if (value is ScalarValue<T>)
+                return ((ScalarValue<T>)value).Value;
```
marrony: Unfortunately not. The only way (I'm aware of) to allow this cast is putting a constraint on method saying that `R` should be a sub-type of `T`, but in that case it won't make sense because most of times we're casting `long` to something else like `char`, `short` or `int`

FaunaDB.Client/Types/Value.cs
marrony: I don't have a strong option about it. I was relying on exceptions the decoder throws. The functional counterpart that receives a codec as you said is fated to disappear, and if I not mistake we created `IResult` for this functional api to not return nulls or throw exceptions. Perhaps we can keep using `IResult` on version 1.0 and remove codecs on version 2.0, what you think?

FaunaDB.Client/Types/Attributes.cs
```diff
@@ -0,0 +1,122 @@
...
+    ///         return new Order
+    ///         {
+    ///             Number = number,
+    ///             Products = products
+    ///         }
+    ///     }
+    /// }
+    /// </example>
+    [AttributeUsage(AttributeTargets.Constructor | AttributeTargets.Method)]
+    public class CreatorAttribute : Attribute
```
marrony: IDK, I was taking a look on [Json.NET attributes ](http://www.newtonsoft.com/json/help/html/serializationattributes.htm) and I started to wonder if we should take step back and prefix our attributes with something like `Fauna`, ex: `FaunaField`, `FaunaIgnore`, `FaunaConstructor`? I'm saying that because just `Field` or `Ignore` already caused some names collision, specially `Field` because we already have a class with that name. Maybe @eaceaser can help us on this

erickpintor: If that is the pattern other libraries are taking, we should also follow. We want C# developers to feel at home. :)

FaunaDB.Client/Types/Decoder.cs
erickpintor: You can't force cast abstract types? Interesting.

FaunaDB.Client/Types/Value.cs
erickpintor: I think there will be always the case where you wan't to traverse the result tree and later on decode it's leaf. For example: 

```csharp
res.Get("key1").Get("key2").To<Product>()
```

In that case, `IResult` comes at hand. Maybe `To` is the final call so, you don't need to return a `IResult` but, at the same time, `IResult` is useful when you are schemaless since you try to decode the value to different types based on their failure instead of `try/catch` conversion attempts.

erickpintor:
Maybe for optional fields? Things that doesn't come from the database but the app could pass along the other parameters at the constructor.

Regardless of its usage, I think it makes our api consistent. I would expect the ignore to also work at the constructor level since the `Field` annotation does.

marrony:
For this case I included `DefaultValue` on `Field` attribute. If the response don't include the respective field, the default value would be used, also it's possible to use the native default value from language like, `Constructor(int value = 10)`, the value `10` would be used case the field `value` is not present on the response.

Anyways the only possible way to ignore a parameter on constructor is having a default value for it, either using `DefaultValue` on `Field` or having a default value on the argument.

FaunaDB.Client/Types/Codec.cs
```diff
@@ -47,6 +47,33 @@ public static IResult<Value> VALUE(Value value)
...
+            {
+                return Success(Decoder.Decode<T>(input));
+            }
+            catch (InvalidOperationException ex)
+            {
+                return Fail<T>(ex.Message);
+            }
+        }
+
+        public static Func<Value, IResult<object>> DECODE(Type dstType)
```
erickpintor: Wound you mind explaining this variation of `DECODE`? What's its use case?

FaunaDB.Client/Types/Decoder.cs
erickpintor: So `R` is only used to build the error message, right? Maybe you can compose with another function to build the error message to avoid the type information in the function signature. It's a little confusing.

FaunaDB.Client/Types/Codec.cs
marrony: I was reviewing this PR and I was thinking to remove that codec but was too late, you started to submit comments 😄 

This was for cases when you don't have the type at compile time, only on runtime. But yes, it might not be correctly pollute the codec api with this. If the user would like to do this they could use the `Decoder` directly.

FaunaDB.Client/Types/Decoder.cs
marrony: I think I don't get you idea, would you mind to provide a sample?

marrony:
I just included test cases for overflow on deserialization

FaunaDB.Client/Types/Decoder.cs
```diff
@@ -0,0 +1,493 @@
...
+
+            return ret;
+        }
+
+        static IDictionary FromDictionary(Value value, Type type)
+        {
+            if (!(value is ObjectV))
+                throw new InvalidOperationException($"Cannot convert `{value}` to {type}");
+
+            if (!type.IsGenericType)
```
eaceaser: Should we have an assertion that type is assignable from `IDictionary`, or equivalent, and that there are only two generic type parameters?

FaunaDB.Client/Types/Decoder.cs
```diff
@@ -0,0 +1,493 @@
...
+
+            throw new InvalidOperationException($"Cannot convert {value} to {dstType}");
+        }
+
+        static IList FromEnumerable(Value value, Type type)
+        {
+            if (!(value is ArrayV))
+                throw new InvalidOperationException($"Cannot convert `{value}` to {type}");
+
+            if (!type.IsGenericType)
```
eaceaser: See the note below in `FromDictionary` - should we assert that the type is assignable from a base interface type?

FaunaDB.Client/Types/Decoder.cs
marrony: This is done in line 131, I wish I could assert that the type is assignable to a generic type, like `IDicitonary<,>` but C# doesn't allow this

FaunaDB.Client/Types/Codec.cs
```diff
@@ -47,6 +47,22 @@ public static IResult<Value> VALUE(Value value)
 
         public static IResult<IReadOnlyDictionary<string, Value>> OBJECT(Value input) =>
             Cast.MapTo<ObjectV, IReadOnlyDictionary<string, Value>>(input, x => x.Value);
+
+        public static IResult<T> DECODE<T>(Value input)
```
marrony: No conventions was used here, it's something I did wrong in the beginning, now we have to live with it or create siblings methods with correct names and deprecate the old ones

FaunaDB.Client/Types/Decoder.cs
marrony: The same case as below, it's tested on line 134

eaceaser: word

PR #42, opened 2017-04-12
Changes from 1.0

marrony:
All the last changes from version 1.0
* serialization of user classes
* deserialization of user classes
* @query type and user functions
* values constructions


erickpintor:
Did you have to change anything or are you just cherry-picking the commits?

marrony:
Sorry, I miss the last comment. The majority of the code is just cherry-pick, I just had to fix a conflict con ref classes because this classes has its own file now.

PR #43, opened 2017-04-12
Release 1.1.0

marrony:

PR #45, opened 2017-04-13
Fix decoder cast among Values type

marrony:
This fix the case when we want to decode a `StringV` to a `Value` for example, since `StringV` extends `Value` we could just assign the string, but instead it was trying to instantiate a `Value`.


marrony:
Need this reviewed before https://github.com/fauna/faunadb-csharp/pull/43

PR #46, opened 2017-04-13
Document (de)serialization of user defined classes

marrony:
close https://github.com/fauna/faunadb-csharp/issues/44

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ Except `secret` all other arguments are optional.
 #### How to execute a query
 
 ```csharp
-Value result = client.Query(Paginate(Match(Ref("indexes/spells"))));
+Value result = await client.Query(Paginate(Match(Ref("indexes/spells"))));
 ```
 
 `Query` methods receives an `Expr` object. `Expr` objects can be composed with others `Expr` to create complex query objects. `FaunaDB.Query.Language` is a helper class where you can find all available expressions in the library.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -154,6 +154,69 @@ IResult<string> result.Map(value => value.toString());
 
 If `result` represents an failure all calls to `Map` and `FlatMap` are ignored. See `FaunaDB.Types.Result`.
 
+### How to work with user defined classes
+
+Instead of create manually your objects using our dsl (via `Obj()` method) it's possible to use our `Encoder` class and convert any user defined class to an equivalent `Value` type.
```
eaceaser: A better rephrasing:

"Instead of manually creating your objects via the DSL (e.g. the `Obj()` method), you may use the `Encoder` class to convert a user-defined type into the equivalent `Value type."

README.md
```diff
@@ -154,6 +154,69 @@ IResult<string> result.Map(value => value.toString());
...
+
+Consider the following `Product` class:
```
eaceaser: Just say: "For example:"

README.md
```diff
@@ -154,6 +154,69 @@ IResult<string> result.Map(value => value.toString());
...
+    [FaunaConstructor]
+    public Product(string description, double price)
+    {
+        Description = description;
+        Price = price;
+    }
+}
+```
+
+It's possible to create and persist an instance of `Product` on FaunaDB using:
```
eaceaser: maybe rephrase: "To persist an instance of `Product` in FaunaDB:"

README.md
```diff
@@ -154,6 +154,69 @@ IResult<string> result.Map(value => value.toString());
...
+
+await client.Query(
+    Create(
+        Ref("classes/product"),
+        Obj("data", Encoder.Encode(product))
+    )
+);
+```
+
+Also it's possible to convert back from a `Value` type to a `Product` type using our `Decoder` class:
```
eaceaser: I'd probably rephrase, dropping the "it's possible":

"To convert from a `Value` type back to the `Product` type, you can use a `Decoder`:"

README.md
```diff
@@ -154,6 +154,69 @@ IResult<string> result.Map(value => value.toString());
...
+
+```csharp
+Value value = await client.Query(Get(Ref("classes/product/123456789")));
+
+Product product = Decoder.Decode<Product>(value);
+```
+
+or using a sugar method `To<T>()` like:
```
eaceaser: "or via the `To<T>()` helper method:"

README.md
```diff
@@ -154,6 +154,69 @@ IResult<string> result.Map(value => value.toString());
...
+
+```csharp
+Value value = await client.Query(Get(Ref("classes/product/123456789")));
+
+IResult<Product> product = value.To<Product>();
+```
+
+In that case it'll return a `IResult<T>` type.
```
eaceaser: "Note that in this case the return type is `IResult<T>`"

README.md
```diff
@@ -154,6 +154,69 @@ IResult<string> result.Map(value => value.toString());
...
+
+Those are three attributes that can be used to change the behavior of the `Encoder` and `Decoder`:
```
eaceaser: "Those are three" => "There are three"

README.md
```diff
@@ -154,6 +154,69 @@ IResult<string> result.Map(value => value.toString());
...
+
+- `FaunaField`: Used to override a custom field name and/or provide a default value for that field. If this attribute is not specified, the member name will be used instead. Can be used on fields, properties and constructors arguments.
```
eaceaser: "constructors" => "constructor"

README.md
```diff
@@ -154,6 +154,69 @@ IResult<string> result.Map(value => value.toString());
...
+- `FaunaConstructor`: Used to mark a constructor or a public static method as an instantiator of the specified type. This attribute can be used only once per class.
```
eaceaser: I think its clearer to change "as an instantiator of the specified type" => "as the method used to instantiate the specified type"

README.md
```diff
@@ -154,6 +154,69 @@ IResult<string> result.Map(value => value.toString());
...
+- `FaunaIgnore`: Used to ignore a specific member. Can be used on fields, properties and constructors arguments. If used on constructor argument, the argument must have a default value.
```
eaceaser: "constructors" => "constructor".

"If used on constructor argument, the argument" => "if used on a constructor argument, that argument"

README.md
```diff
@@ -154,6 +154,69 @@ IResult<string> result.Map(value => value.toString());
...
+
+All primitive types are supported as data members for your classes including primitive arrays and generic collections types like `List<T>` or `Dictionary<string, T>`. When decoding, if your members use the generic interfaces `IList<T>` or `IDictionary<string, T>` the concrete versions `List<T>` and `Dictionary<string, T>` will be instantiated respectively.
```
eaceaser: I'd simplify this into a list maybe:

===

`Encoder` and `Decoder` can currently convert:

* Primitive scalar types (`int`, `long`, etc.)
* Primitive arrays, generic collections such as `List<T>`, and their respective interfaces such as `IList<T>`.
* Dictionaries with string keys, such as `Dictionary<string, T>` and its respective interface `IDictionary<string, T>`.

===
Does that seem better?

Issue #49, opened 2017-05-15
Support Enum 


erickpintor:
User's feedback:

> C# driver: Noticing that enums are getting encoded as int's. Would it be possible to configure the Encoder to use the string value for an enum instead?
This makes it much easier when quering using the Query Console.

Issue #48, opened 2017-05-13
Having null properties on Entities without using FaunaField attribute makes Decoder.Decode throw exception


overgaardmorten:
Save instance of User without filling out one of FirstName, LastName or Login makes Decoder.Decode<T> throw exception with message: Missing required property: PropertyName

```csharp 
public class User 
{
        public string FirstName { get; set; }
        public string LastName { get; set; }
        public Login Login { get; set; }
    }

    public class Login
    {
        public LoginTypeEnum LoginType { get; set; }

        public string Username { get; set; }
        public string PasswordHash { get; set; } 
        public string FacebookToken { get; set; }
    }
```

It would be nice if Decoder handles properties with null values simply by setting these to null when calling Decode.

marrony:
@overgaardmorten I have a pr https://github.com/fauna/faunadb-csharp/pull/50 ready to review, when approved I should release a new version of the driver

PR #50, opened 2017-05-16
Post launch feedbacks

marrony:
It includes some user requests/feedbacks and a minor fix when decoding value types (aka structs)

close https://github.com/fauna/faunadb-csharp/issues/48
close https://github.com/fauna/faunadb-csharp/issues/49


erickpintor:
Just as a note: this fixes should go to version 2.0 too, I guess.

PR #59, opened 2017-12-06
Split code in two modules

marrony:
depends on:

https://github.com/fauna/faunadb-csharp-api/pull/1
https://github.com/fauna/faunadb-csharp-common/pull/1


erickpintor:
I'm going to look at the PR's soon but I have a question first: is it possible to keep all the code in the same repository? I feel like multiple repos increase the complexity of the driver's development cycle.

marrony:
I don't see how because we would have to control 3 projects. If we need to fix a bug in one module we would have to checkout that module in a specific branch, if we had only one repo, we wouldn't be able to do that.

erickpintor:
I think this is a important discussion. If we break drivers in multiple repos, than we increase exponentially the number of versions we have to keep compatible. Also, it's harder to test because we then need some sort of test suite per LTS to ensure that all modules' versions are working fine together.

I personally feel like a monorepo approach would be simpler to maintain. In that case, we'd split the code in different modules and always use the latest version of each module when releasing. That'd require a api module per core api version. The LTS would be a module which includes the proper sub-modules for that LTS version. Core uses a similar approach to be able to release multiple artifacts out of the same code base.

The downside is an explosion of driver's api code (which will be mostly auto-generated). But at the bright side, we only manage a single repository and no need to fix bugs in old branches.

@eaceaser  might have opinions about this too.

PR #70, opened 2018-07-12
Add conversion functions.

bitbckt:
`ToString()` is C#'s version of Java's `toString()`, so we cannot override it in a static class.


marrony:
yes, it's in `ClientTest.cs`

PR #71, opened 2018-07-27
Support X-Last-Seen-Txn.

bitbckt:
Once again, I don't entirely know what I'm doing here.

### DIFF
--- a/FaunaDB.Client/Client/DefaultClientIO.cs
+++ b/FaunaDB.Client/Client/DefaultClientIO.cs
@@ -20,18 +20,21 @@ namespace FaunaDB.Client
         readonly HttpClient client;
         readonly AuthenticationHeaderValue authHeader;
 
-        internal DefaultClientIO(HttpClient client, AuthenticationHeaderValue authHeader)
+        private LastSeen lastSeen;
+
+        internal DefaultClientIO(HttpClient client, AuthenticationHeaderValue authHeader, LastSeen lastSeen)
         {
             this.client = client;
             this.authHeader = authHeader;
+            this.lastSeen = lastSeen;
         }
 
         public DefaultClientIO(string secret, Uri endpoint, TimeSpan timeout)
-            : this(CreateClient(endpoint, timeout), AuthHeader(secret))
+            : this(CreateClient(endpoint, timeout), AuthHeader(secret), new LastSeen())
         { }
 
         public IClientIO NewSessionClient(string secret) =>
-            new DefaultClientIO(client, AuthHeader(secret));
+            new DefaultClientIO(client, AuthHeader(secret), lastSeen);
 
         public Task<RequestResult> DoRequest(HttpMethodKind method, string path, string data, IReadOnlyDictionary<string, string> query = null) =>
             DoRequestAsync(method, path, data, query);
@@ -49,6 +52,11 @@ namespace FaunaDB.Client
             message.Content = dataString;
             message.Headers.Authorization = authHeader;
 
+            var last = lastSeen.Txn;
+            if (last.HasValue) {
+                message.Headers.Add("X-Last-Seen-Txn", last.Value.ToString());
+            }
+
             var httpResponse = await client.SendAsync(message).ConfigureAwait(false);
 
             string response;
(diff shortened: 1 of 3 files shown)


FaunaDB.Client/Client/FaunaClient.cs
```diff
@@ -122,6 +122,13 @@ public FaunaClient(IClientIO clientIO)
 
             RaiseForStatusCode(responseHttp);
 
+            if (responseHttp.ResponseHeaders.ContainsKey("X-Txn-Time")) {
+                // there shouldn't ever be more than one...
```
marrony: what?

FaunaDB.Client/Client/DefaultClientIO.cs
```diff
@@ -49,6 +49,11 @@ public DefaultClientIO(string secret, Uri endpoint, TimeSpan timeout)
             message.Content = dataString;
             message.Headers.Authorization = authHeader;
 
+            var lastSeen = LastSeen.Txn;
```
marrony: what if you make `LastSeen` a class variable....

FaunaDB.Client/Client/LastSeen.cs
```diff
@@ -0,0 +1,46 @@
+using System;
+using System.Threading;
+
+/// <summary>
+///   A thread-safe static helper used to store the last seen
+///   transaction time returned by the database.
+/// </summary>
+namespace FaunaDB.Client {
+    public class LastSeen {
+        private static long lastSeenTxn = 0;
```
marrony: and remove the static qualifier?

FaunaDB.Client/Client/FaunaClient.cs
bitbckt: More than one X-Txn-Time. Too vague?

FaunaDB.Client/Client/DefaultClientIO.cs
bitbckt: That would share per client, but not across clients. The expected concurrency in these clients isn't clear to me from one to another, so I decided to make this global. Is per-client "better"?

FaunaDB.Client/Client/FaunaClient.cs
marrony: doesn't `responseHttp.ResponseHeaders` represent all the response headers?

FaunaDB.Client/Client/DefaultClientIO.cs
bitbckt: (I think I have that right... C# is also not at all my thing)

marrony: well, given this is a library, make things global might make the users code unpredictable

FaunaDB.Client/Client/FaunaClient.cs
bitbckt: Hmm. I interpreted the type to be a dictionary with string -> enum[string]...

bitbckt: .NET apparently supports multiple-value headers? Weird. https://msdn.microsoft.com/en-us/library/system.net.http.headers.httpheaders(v=vs.118).aspx

FaunaDB.Client/Client/DefaultClientIO.cs
```diff
@@ -20,22 +20,28 @@ class DefaultClientIO : IClientIO
...
         { }
 
         public IClientIO NewSessionClient(string secret) =>
-            new DefaultClientIO(client, AuthHeader(secret));
+            new DefaultClientIO(client, AuthHeader(secret), lastSeen);
 
         public Task<RequestResult> DoRequest(HttpMethodKind method, string path, string data, IReadOnlyDictionary<string, string> query = null) =>
             DoRequestAsync(method, path, data, query);
 
+        public void UpdateLastSeen(long time) =>
```
erickpintor: Can this be private or internal?

FaunaDB.Client/Client/LastSeen.cs
```diff
@@ -0,0 +1,46 @@
...
+            {
+                initial = lastSeenTxn;
+
+                // don't move the last seen time backward
+                if (time <= initial) {
+                    return;
+                }
+            } while (initial != Interlocked.CompareExchange(ref lastSeenTxn, time, initial));
+
+            return;
```
erickpintor: Do you need this return?

bitbckt: Apparently not. Java-ism, I guess.

FaunaDB.Client/Client/DefaultClientIO.cs
bitbckt: It's an interface method, called from the client - I think it needs to be public?

marrony: you don't need to expose this method on the interface, you can do all the job on `DefaultClientIO`

bitbckt: Aha. So I can. Coming right up.

bitbckt:
I'm using `csharp-mode`. ;-)

PR #81, opened 2019-11-05
Add Numeric/Math functions

evbruno:


FaunaDB.Client/Query/Language.Miscellaneous.cs
```diff
@@ -439,5 +439,303 @@ public partial struct Language
...
+        public static Expr Trunc(Expr expr) =>
+          UnescapedObject.With("trunc", expr);
+
+        /// <summary>
+        /// Creates a new Trunc expression.
+        /// <para>
+        /// See the <see href="https://docs.fauna.com/fauna/current/api/fql/functions/trunc">FaunaDB Trunc Function</see>
+        /// </para>
+        /// </summary>
+        public static Expr Trunc(Expr expr, Expr precision) =>
```
marrony: you can remove the above overload and leave just this one with `Expr precision = null`

PR #85, opened 2019-12-05
[ENG-1512] Add Now function

evbruno:


FaunaDB.Client.Test/ClientTest.cs
```diff
@@ -1692,6 +1692,14 @@ public async Task TestStringFunctions()
             Assert.AreEqual(StringV.Of("HELLO WORLD"), await client.Query(UpperCase("hello world")));
         }
 
+        [Test]
+        public async Task TestNowFunction()
+        {
+            DateTime systemDate = DateTime.Now.ToUniversalTime();
+            DateTime now = (await client.Query(Now())).To<DateTime>().Value;
+            Assert.IsTrue(systemDate.Second - now.Second <= 1);
```
marrony: this might not always be true given the db time is different from the client issuing the query, use a db-only approach
https://github.com/fauna/faunadb-jvm/blob/master/faunadb-java/src/test/java/com/faunadb/client/ClientSpec.java#L1709

evbruno:
fixed

PR #86, opened 2019-12-06
[ENG-1515] add functions Sum, Mean and Count

evbruno:


FaunaDB.Client/Query/Language.Read.cs
```diff
@@ -72,5 +72,32 @@ public partial struct Language
         public static Expr Reduce(Func<Expr, Expr, Expr> lambda, Expr initial, Expr collection) =>
             Reduce(Lambda(lambda), initial, collection);
 
+        /// <summary>
+        /// Returns the number of items that exist in the array or set
+        /// <para>
+        /// See the <see href="https://docs.fauna.com/fauna/current/api/fql/functions/count">FaunaDB Count Function</see>
+        /// </para>
+        /// </summary>
+        public static Expr Count(params Expr[] terms) =>
```
marrony: I don't think vargars would be a good idea, calling `Count(1, 2, 3)` is nice, but it would allow `Count(Match(...), Match(...))` which is not allowed

https://github.com/fauna/faunadb-jvm/blob/master/faunadb-java/src/main/java/com/faunadb/client/query/Language.java#L6627

FaunaDB.Client/Query/Language.TimeDate.cs
```diff
@@ -51,5 +51,14 @@ public enum TimeUnit
...
         public static Expr Date(Expr date) =>
             UnescapedObject.With("date", date);
+
+        /// <summary>
+        /// Constructs a Timestamp representing the transaction’s start time.
+        /// <para>
+        /// See the <see href="https://docs.fauna.com/fauna/current/api/fql/functions/now">FaunaDB Now Functions</see>
+        /// </para>
+        /// </summary>
+        public static Expr Now() =>
```
marrony: I think this doesn't belong to this PR :D

evbruno: properly rebased 🤦🏻‍♂️

PR #88, opened 2020-01-19
Update README.md

iwatakeshi:
Hi, the following are some fixes to the documentation:

* Replaces`Class` with `Collection`
* Decoding example should use `.At("data")` to properly decode the value.

Let me know if there are any mistakes or changes needed.

Thank you.


evbruno:
nice catch, thanks !

PR #90, opened 2020-03-18
Feature to enforce DateV and TimeV for encoder plus DateTimeExtensions

kcrandall:
Spoke to Marrony Neris on community slack about the issue with Time equal to zero always causing DateTime to be converted into DateV even if you want TimeV

### DIFF
--- a/FaunaDB.Client/Types/Attributes.cs
+++ b/FaunaDB.Client/Types/Attributes.cs
@@ -49,13 +49,22 @@ namespace FaunaDB.Types
         /// The default value used for a missing property when decoding an object.
         /// </summary>
         public object DefaultValue { get; set; }
-
+        
         public FaunaFieldAttribute(string name)
         {
             Name = name;
         }
     }
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property | AttributeTargets.Parameter)]
+    public class FaunaDate : Attribute
+    {
 
+    }
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property | AttributeTargets.Parameter)]
+    public class FaunaTime : Attribute
+    {
+
+    }
     /// <summary>
     /// Instruct the encoder to not encode the specified member.
     /// </summary>
--- a/FaunaDB.Client/Types/Encoder.cs
+++ b/FaunaDB.Client/Types/Encoder.cs
@@ -46,7 +46,7 @@ namespace FaunaDB.Types
 
         Stack<object> stack = new Stack<object>();
 
-        public Value Encode(object obj)
+        public Value Encode(object obj, Type t = null)
         {
             if (obj == null)
                 return NullV.Instance;
@@ -61,7 +61,7 @@ namespace FaunaDB.Types
             {
                 stack.Push(obj);
 
-                return EncodeIntern(obj);
+                return EncodeIntern(obj, t);
             }
             finally
             {
@@ -81,7 +81,7 @@ namespace FaunaDB.Types
 #endif
         }
 
-        Value EncodeIntern(object obj)
+        Value EncodeIntern(object obj, Type forceType)
         {
             var type = obj.GetType();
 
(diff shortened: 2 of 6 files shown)


FaunaDB.Client/Types/Value.cs
```diff
@@ -164,16 +164,26 @@ public abstract class Value : Expr
...
 
-        internal static Value FromDateTime(DateTime dt)
+        internal static Value FromDateTime(DateTime dt, Type t = null)
         {
+            if(t == typeof(TimeV))
+                return new TimeV(dt);
+            else if(t == typeof(DateV))
+                return new DateV(dt);
+
             if (dt.Ticks % (24 * 60 * 60 * 10000) > 0)
```
marrony: instead of the above logic, you can just add a `||` condition on this `if`
```csharp
if(t == typeof(TimeV) || dt.Ticks % (24 * 60 * 60 * 10000) > 0)
```

FaunaDB.Client/Types/Value.cs
```diff
@@ -164,16 +164,26 @@ public abstract class Value : Expr
...
 
-        internal static Value FromDateTimeOffset(DateTimeOffset dt)
+        internal static Value FromDateTimeOffset(DateTimeOffset dt, Type t = null)
         {
+            if (t == typeof(TimeV))
+                return new TimeV(dt.UtcDateTime);
+            else if (t == typeof(DateV))
+                return new DateV(dt.UtcDateTime);
+
             if (dt.UtcTicks % (24 * 60 * 60 * 10000) > 0)
```
marrony: same here

FaunaDB.Client/Types/Encoder.cs
```diff
@@ -100,7 +100,7 @@ Value EncodeIntern(object obj)
                     return BooleanV.Of((bool)obj);
 
                 case TypeCode.DateTime:
-                    return Value.FromDateTime((DateTime)obj);
+                    return Value.FromDateTime((DateTime)obj, t);
```
marrony: you need to pass `t` argument to `Value.FromDateTimeOffset()` a few lines below

FaunaDB.Client/Types/Encoder.cs
```diff
@@ -81,7 +81,7 @@ static object GetDBNullValue()
 #endif
         }
 
-        Value EncodeIntern(object obj)
+        Value EncodeIntern(object obj, Type t)
```
marrony: rename `t` to something less terse like `type`

FaunaDB.Client/Types/Attributes.cs
```diff
@@ -50,10 +50,35 @@ public class FaunaFieldAttribute : Attribute
         /// </summary>
         public object DefaultValue { get; set; }
 
+        /// <summary>
+        /// Overrides the fauna type <see cref="FaunaDB.Types"/> namespace for
+        /// list of accepted values.
+        /// Type should inherit from <see cref="FaunaDB.Types.Value"/>.
+        /// Such as, <see cref="TimeV"/>, <see cref="DateV"/>, <see cref="DoubleV"/>, etc
+        ///
+        /// Not fully supprted, only use for the <see cref="Encoder"/> specifically to enforce
```
marrony: I'm not liking too much of this, isn't better to create a dedicated attribute just for this? Like `FaunaDateAttribute/FaunaTimeAttribute`?

FaunaDB.Client/Types/Attributes.cs
```diff
@@ -50,10 +50,35 @@ public class FaunaFieldAttribute : Attribute
...
+        /// DateV or TimeV for <see cref="DateTime"/> and <see cref="DateTimeOffset"/> properties
+        /// </summary>
+        public Type Type { get; }
+
         public FaunaFieldAttribute(string name)
         {
             Name = name;
         }
+        public FaunaFieldAttribute(string name, Type type)
```
marrony: Are those combination of constructor really necessary? I wonder if we add a new field to this attribute how exponential this will be. I think this kinda reinforce the my argument above.

Issue #73, opened 2019-05-14
HttpClient & TCP Connections


mayoatte:
Hey folks,

Just came across FaunaDb. While checking out this driver I noticed that it seems to create a new HttpClient every time.

A lot of the serverless platforms (FaaS ones like Lambda and Azure Functions) have [limits on the number of connections](https://docs.microsoft.com/en-us/azure/azure-functions/manage-connections) that can be created inside a function and inappropriate usage HttpClient is frequently the cause of many errors.

The old way to fix this was to use a static http client but that has problems as well. The newer way is to use an [HttpClientFactory](https://docs.microsoft.com/en-us/dotnet/standard/microservices-architecture/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests) which does a much better job of managing connections.

Anyways. Just wanted to bring that up. The serverless tier of FaunaDb is what caught my eye and this is a frequent gotcha for those scenarios.

kcrandall:
evbruno this can be closed now

PR #93, opened 2020-03-29
DRV-59: Added ability to use custom HttpClient - fixes issue #73 

kcrandall:
See issue #73 for link as to why this is important. Simple fix to allow library user to supply their own HttpClient


FaunaDB.Client/Client/DefaultClientIO.cs
```diff
@@ -29,9 +32,12 @@ internal DefaultClientIO(HttpClient client, AuthenticationHeaderValue authHeader
             this.lastSeen = lastSeen;
         }
 
-        public DefaultClientIO(string secret, Uri endpoint, TimeSpan timeout)
-            : this(CreateClient(endpoint, timeout), AuthHeader(secret), new LastSeen())
-        { }
+        public DefaultClientIO(string secret, Uri endpoint, TimeSpan timeout, HttpClient httpClient = null)
+            : this(CreateClient(endpoint, timeout, httpClient), AuthHeader(secret), new LastSeen())
```
marrony: no need to pass `httpClient` to the function `CreateClient`, if you already have it, just ignore the function call.
```csharp
httpClient ?? CreateClient(endpoint, timeout)
```

FaunaDB.Client/Client/DefaultClientIO.cs
```diff
@@ -117,17 +128,16 @@ static string QueryString(IReadOnlyDictionary<string, string> query)
             return string.Join("&", keyValues);
         }
 
-        static HttpClient CreateClient(Uri endpoint, TimeSpan timeout)
+        static HttpClient CreateClient(Uri endpoint, TimeSpan timeout, HttpClient httpClient)
```
marrony: remove `HttpClient httpClient`

FaunaDB.Client/Client/DefaultClientIO.cs
```diff
@@ -48,9 +54,14 @@ async Task<RequestResult> DoRequestAsync(HttpMethodKind method, string path, str
 
             var startTime = DateTime.UtcNow;
 
-            var message = new HttpRequestMessage(new HttpMethod(method.Name()), path);
+            var message = new HttpRequestMessage(new HttpMethod(method.Name()), $"{endpoint}{path}");
```
marrony: why did you add `endpoint` here? the client already has the endpoint as the `BaseAddress`

FaunaDB.Client/Client/DefaultClientIO.cs
```diff
@@ -48,9 +54,14 @@ async Task<RequestResult> DoRequestAsync(HttpMethodKind method, string path, str
...
             message.Content = dataString;
             message.Headers.Authorization = authHeader;
+            message.Headers.AcceptEncoding.Add(new StringWithQualityHeaderValue("gzip"));
+            message.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+            message.Headers.Add("X-FaunaDB-API-Version", "2.7");
+            message.Headers.Add("X-Fauna-Driver", "csharp");
+            message.SetTimeout(timeout);
```
marrony: why are you setting timeout here? we are already setting this on `HttpClient` creation

FaunaDB.Client/Client/DefaultClientIO.cs
kcrandall: If the user is passing their own then they might not have a base address set and you don't want to change theirs or set one

FaunaDB.Client/Client/DefaultClientIO.cs
kcrandall: I removed that, because the user may pass their own. So you need to set it on the HttpRequestMessage for each call. The user may have their own timeout or the default. Don't want to change it and mess up how their other requests deal with them

FaunaDB.Client/Client/DefaultClientIO.cs
kcrandall: I need to pass the httpClient if the user passes their own in FaunaClient(httpClient: userDefinedClient)

FaunaDB.Client/Client/DefaultClientIO.cs
kcrandall: have to pass it from FaunaClient() constructor

FaunaDB.Client/Client/DefaultClientIO.cs
kcrandall: wait I see what you are saying on this one

kcrandall: Ya we can ignore the call too I guess

FaunaDB.Client/Client/DefaultClientIO.cs
kcrandall: and this can change as a result of the last one

FaunaDB.Client/Client/DefaultClientIO.cs
marrony: can you add a test where you create your own `HttpClient` not specifying `BaseAddress` on it, just to make sure we are handling the `endpoint` + `path` correctly

kcrandall: sure what test file do you want it in?

kcrandall: ClientTest.cs?

kcrandall: How do I run unit tests that require networking? they say connection refused do I need to supply a secret key somewhere?

kcrandall: The client should act the same though with user supplied or without because even if the client creates its own its not setting a base. So even without user supplied httpclient if it doesn't work all the current tests will fail

marrony: pretty much all tests extends this https://github.com/fauna/faunadb-csharp/blob/master/FaunaDB.Client.Test/TestCase.cs#L42 so you can set an env variable with your key/secret and domain/port.

Anyway, without a test I cannot ensure that a custom httpclient set with a BaseAddress will not interfere with endpoint+path

FaunaDB.Client/Client/DefaultClientIO.cs
```diff
@@ -120,12 +131,8 @@ static string QueryString(IReadOnlyDictionary<string, string> query)
         static HttpClient CreateClient(Uri endpoint, TimeSpan timeout)
         {
             var client = new HttpClient();
-            client.BaseAddress = endpoint;
-            client.Timeout = timeout;
-            client.DefaultRequestHeaders.AcceptEncoding.Add(new StringWithQualityHeaderValue("gzip"));
-            client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
-            client.DefaultRequestHeaders.Add("X-FaunaDB-API-Version", "2.7");
-            client.DefaultRequestHeaders.Add("X-Fauna-Driver", "csharp");
+            //client.BaseAddress = endpoint;
```
marrony: remove commented code

FaunaDB.Client/Client/DefaultClientIO.cs
```diff
@@ -47,10 +53,15 @@ async Task<RequestResult> DoRequestAsync(HttpMethodKind method, string path, str
                 path = $"{path}?{queryString}";
 
             var startTime = DateTime.UtcNow;
-
-            var message = new HttpRequestMessage(new HttpMethod(method.Name()), path);
+            System.Diagnostics.Debug.WriteLine($"Endpoint: {endpoint}{path}");
```
marrony: don't forget to remove this

FaunaDB.Client/Client/DefaultClientIO.cs
marrony: Or you can set the base address of the custom httpclient to a non related address and ensure we are doing requests to fauna address specified on endpoint

kcrandall: Ok i passed an httpclient in TestCase with a random baseaddress. it worked as expected and the tests passed. HttpRequestMessage will always override the base address if a domain is passed to it

evbruno:
followup ticket: DRV-99

FaunaDB.Client/Client/FaunaClient.cs
```diff
@@ -33,8 +33,9 @@ public class FaunaClient
         public FaunaClient(
             string secret,
             string endpoint = "https://db.fauna.com:443",
-            TimeSpan? timeout = null)
-            : this(CreateClient(secret, endpoint, timeout))
+            TimeSpan? timeout = null,
+            System.Net.Http.HttpClient httpClient = null)
```
marrony, line 37: could you add the import for this class in the top of the file and remove the full qualified name here?

FaunaDB.Client/Client/HttpRequestExtensions.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.Net.Http;
+namespace FaunaDB.Client
+{
+    public static class HttpRequestExtensions
+    {
+        private static string TimeoutPropertyKey = "RequestTimeout";
```
marrony, line 7: Where did you find oficial documentation about this property? Microsoft documentation don't mention anything about it https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httprequestmessage.properties?view=netframework-4.8

kcrandall: https://thomaslevesque.com/2018/02/25/better-timeout-handling-with-httpclient/

kcrandall: this is another alternative to set cancellation token https://stackoverflow.com/questions/28060575/how-to-set-time-out-for-http-client-request-operation-in-windows-phone-8-1-windo

kcrandall: I can add the cancellation token too if that's better

marrony: looks like `request.Properties` is not used in any place on the source code https://github.com/microsoft/referencesource/tree/master/System/net/System/Net/Http

kcrandall: looks like the best thing to do is this. It will cancel the request when the token expires causing it timeout 
```
CancellationTokenSource cts = new CancellationTokenSource(2000); // 2 seconds
    HttpClient client = new HttpClient();
    HttpResponseMessage response = await
        client.SendRequestAsync(request).AsTask(cts.Token);
```

evbruno: I think we could remove the extension, and add the timeout in the client creation... I can confirm that this patch works:
```
diff --git a/FaunaDB.Client/Client/DefaultClientIO.cs b/FaunaDB.Client/Client/DefaultClientIO.cs
index 7e1f3b7..0e5fd0c 100644
--- a/FaunaDB.Client/Client/DefaultClientIO.cs
+++ b/FaunaDB.Client/Client/DefaultClientIO.cs
@@ -35,7 +35,7 @@ namespace FaunaDB.Client
         }

         public DefaultClientIO(string secret, Uri endpoint, TimeSpan timeout, HttpClient httpClient = null)
-            : this(httpClient ?? CreateClient(), AuthHeader(secret), new LastSeen(), endpoint, timeout)
+            : this(httpClient ?? CreateClient(timeout), AuthHeader(secret), new LastSeen(), endpoint, timeout)
         { }

@@ -127,9 +127,14 @@ namespace FaunaDB.Client
             return string.Join("&", keyValues);
         }

-        static HttpClient CreateClient()
+        static HttpClient CreateClient(TimeSpan? timeout)
         {
-            return new HttpClient();
+            var c = new HttpClient();
+
+            if (timeout.HasValue)
+                c.Timeout = timeout.Value;
+
+            return c;
         }
     }
 }```

evbruno: Ok... you can remove every timeout related code, we are going to manage this in the server side.

kcrandall: I assume you'll need to send timeout in some header or something so ill just remove the extensions and SetTimeout() call in DefaultClientIO and leave timeout in the DefaultClientIO constructor?

marrony: yeah, that's the idea, the driver will send a header with the desired timeout per query and it will control the entire pipeline on the server side

kcrandall: ok so the new commit should fix that you guys just have to add the header in a separate commit

kcrandall: if you want to give me the header key I can set it to the timeout variable

evbruno: Before we move to this new header approach, I would like to finish and merge this PR. Could  you do it ? We just need to remove the current timeout related stuff and this PR is good to go!

evbruno: In the meantime I'm looking for the header specs (such as, is this milliseconds or seconds?)

kcrandall: Wouldn't the best way to do it server side is have two headers one is an int. The other is an enum with values mil, sec, tick etc. Then it doesn't matter what the clients send users can send whatever they want and server side figures out the conversions

kcrandall: I don't off the top of my head know what C# timeout is bc I always leave it default

kcrandall: I removed the timeout stuff too in commit f0a1c92. It still exists in the class ready to be put in a header

kcrandall: the client uses TimeSpan so users can set it to anything TimeSpan.FromSeconds()/Hours() etc

kcrandall: you can do this on the client timeout.TotalMilliseconds to then pass it to the header

evbruno: I'm not sure YET how the header is going to behave but using TimeSpan to convert to milliseconds or seconds sounds really nice (user friendly)

PR #94, opened 2020-03-31
DRV-58: Encode raw exprs

evbruno:


FaunaDB.Client.Test/SerializationTest.cs
```diff
@@ -1090,5 +1091,26 @@ public void TestDocuments()
...
+                { "source", Collection("class_name") }
+            };
+
+            AssertJsonEqual(
+                Encode(indexCfg),
+                "{\"object\":{\"name\":\"index_name\",\"source\":{\"collection\":\"class_name\"}}}"
+           );
+
+            AssertJsonEqual(
+                CreateIndex((Expr)indexCfg),
```
marrony, line 1110: I don't remember, but does the cast `(Expr)` really necessary?

evbruno: you're write... we don't need that cast

PR #95, opened 2020-03-31
DRV-54: Paginate cursor

evbruno:


evbruno:
blocked by #94

FaunaDB.Client/Query/Language.Read.cs
```diff
@@ -35,15 +38,55 @@ public partial struct Language
...
+                { "paginate", set ?? NullV.Instance },
+                { "ts", ts },
+                { "size", size },
+                { "events", events },
+                { "sources", sources }
+            };
+
+            if (cursor != null)
+                exprs.Add("cursor", cursor.Source);
+            else if (after != null)
```
marrony, line 55: I think you can remove the `else` clause, because there is no order of preference here. If the user send all cursors at the same time it would be an error anyway.

evbruno: ok 👍

FaunaDB.Client/Query/Unescaped.cs
```diff
@@ -63,6 +63,7 @@ public override string ToString()
 
         public static UnescapedObject With(string key1, Expr value1, string key2, Expr value2, string key3, Expr value3, string key4, Expr value4, string key5, Expr value5, string key6, Expr value6, string key7, Expr value7) =>
             new UnescapedObject(ImmutableDictionary.Of(key1, value1 ?? NullV.Instance, key2, value2, key3, value3, key4, value4, key5, value5, key6, value6, key7, value7));
+
```
marrony, line 66: just revert this file

PR #97, opened 2020-04-23
DRV-85: query timeout header

evbruno:


FaunaDB.Client.Test/TestCase.cs
```diff
@@ -98,4 +97,19 @@ public MockClientIO(RequestResult resp)
         public Task<RequestResult> DoRequest(HttpMethodKind method, string path, string data, IReadOnlyDictionary<string, string> query = null) =>
             Task.FromResult(resp);
     }
+
+    class HttpClientWrapper : HttpClient
+    {
+
+        HttpRequestMessage LastMessage;
```
marrony, line 104:
```csharp
public HttpRequestMessage LastMessage { get; private set; }
```
if I not mistake

FaunaDB.Client.Test/TestCase.cs
```diff
@@ -40,21 +42,18 @@ async Task SetUpAsync()
...
             var port = Env("FAUNA_PORT", "8443");
-            var secret = Env("FAUNA_ROOT_KEY", "secret");
-            var endpoint = $"{scheme}://{domain}:{port}";
-            var httpClient = new HttpClient();
-            httpClient.BaseAddress = new Uri("https://www.wrongbaseaddress.com");
 
-            rootClient = new FaunaClient(secret: secret, endpoint: endpoint, httpClient: httpClient);
+            faunaSecret = Env("FAUNA_ROOT_KEY", "secret");
+            faunaEndpoint = $"{scheme}://{domain}:{port}";
+            rootClient = new FaunaClient(secret: faunaSecret, endpoint: faunaEndpoint, httpClient: new HttpClient());
```
marrony, line 48: aren't the driver already instantiating `HttpClient` if we do not provide one? if so, you can just remove this argument

evbruno: fixed

evbruno:
updated the `query` method, README file... good for further reviewing

Issue #96, opened 2020-04-17
Feature Request: Encoder/Decoder Uri and similar types as a string


kcrandall:
would be nice if you could have a Uri be encoded as a string into faunadb. this can be accomplished similar to the [FaunaDate] and [FaunaTime] attributes. You can have a [FaunaString] that encodes the object by calling the .ToString() and the decoder calls a constructor with a single parameter, string.

evbruno:
issue merged on PR #99

PR #99, opened 2020-04-27
Issue #96 to add [FaunaString] attribute

kcrandall:
The goal with this is to allow people to add the FaunaString attr to their properties and it will encode them using the ToString() method into StringV. The decoder will call the constructor of the object that accepts string.

### DIFF
--- a/FaunaDB.Client.Test/EncoderTest.cs
+++ b/FaunaDB.Client.Test/EncoderTest.cs
@@ -369,5 +369,33 @@ namespace Test
                 Encode(indexCfg)
             );
         }
+
+        class StringOverride
+        {
+            [FaunaField("uri")]
+            [FaunaString]
+            public Uri Uri { get; set; }
+
+            [FaunaField("guid")]
+            [FaunaString]
+            public Guid Guid { get; set; }
+
+            public StringOverride(Uri uri, Guid guid)
+            {
+                this.Uri = uri;
+                this.Guid = guid;
+            }
+        }
+
+        [Test]
+        public void TestStringOverride()
+        {
+            Uri testUri = new Uri("https://fauna.com");
+            Guid testGuid = Guid.NewGuid();
+            Assert.AreEqual(
+                ObjectV.With("uri", new StringV(testUri.ToString()), "guid", new StringV(testGuid.ToString())),
+                Encode(new StringOverride(testUri, testGuid))
+            );
+        }
     }
 }
--- a/FaunaDB.Client/Types/Attributes.cs
+++ b/FaunaDB.Client/Types/Attributes.cs
@@ -92,7 +92,31 @@ namespace FaunaDB.Types
     /// </example>
     [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property | AttributeTargets.Parameter)]
     public class FaunaTime : Attribute
-    {}
+    { }
+
+    /// <summary>
+    /// Instruct the encoder that this object should be treated as a string when
+    /// stored in Fauna. The Encoder will call the object or primative type's
+    /// .ToString() method. The Decoder will attempt to create an object with the
+    /// constructor with a single string parameter. If it does not have that
+    /// constructor the Decoder will fail. For primatives the Decoder will attempt
+    /// to use standard system conversions to convert the string back to the
+    /// primative type.
+    /// </summary>
+    /// <example>
+    /// class Website
+    /// {
+    ///     [FaunaString]
+    ///     public Uri MyLink { get; set; }
+    /// }
+    ///
+    /// var web = new Website { MyLink = new Uri("http://fauanadb.com") };
+    ///
+    /// var encodedWebsite = Encoder.Encode(web);
+    /// </example>
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property | AttributeTargets.Parameter)]
+    public class FaunaString : Attribute
+    { }
 
     /// <summary>
     /// Instruct the encoder to not encode the specified member.
--- a/FaunaDB.Client/Types/Decoder.cs
+++ b/FaunaDB.Client/Types/Decoder.cs
@@ -72,7 +72,7 @@ namespace FaunaDB.Types
             return new InvalidCastException($"Invalid cast from '{from}' to '{to}'.");
         }
 
-        public static object Decode(Value value, Type dstType)
+        public static object Decode(Value value, Type dstType, Type forceType = null)
         {
             if (value == NullV.Instance && dstType == typeof(Value))
                 return value;
@@ -80,10 +80,10 @@ namespace FaunaDB.Types
             if (value == null || value == NullV.Instance)
                 return dstType.GetTypeInfo().IsValueType ? Activator.CreateInstance(dstType) : null;
 
-            return DecodeIntern(value, dstType);
+            return DecodeIntern(value, dstType, forceType);
         }
 
-        static object DecodeIntern(Value value, Type dstType)
+        static object DecodeIntern(Value value, Type dstType, Type forceType = null)
         {
             if (typeof(Value).IsAssignableFrom(dstType))
             {
@@ -143,7 +143,17 @@ namespace FaunaDB.Types
 
                 case TypeCode.Object:
                     if (typeof(DateTimeOffset) == dstType)
+                    {
+                        if (forceType == typeof(StringV) && value.GetType() == typeof(StringV))
+                        {
+                            if (value != null && value != NullV.Instance)
+                                return new DateTimeOffset(Convert.ToDateTime(ToString(value)));
+                            else if (dstType.Is(typeof(Nullable<>)))
+                                return null;
+                            return default(DateTimeOffset);
+                        }
                         return new DateTimeOffset(ToDateTime(value));
+                    }
 
                     if (typeof(byte[]) == dstType && value is BytesV)
                         return ToByteArray(value);
@@ -159,6 +169,14 @@ namespace FaunaDB.Types
 
                     if (typeof(IList).IsAssignableFrom(dstType) || (dstType.GetTypeInfo().IsGenericType && dstType.Is(typeof(IEnumerable<>))))
                         return FromEnumerable(value, dstType);
+
+                    // All other objects with forcetype StringV
+                    if (forceType == typeof(StringV) && value.GetType() == typeof(StringV))
+                    {
+                        if (value != null && value != NullV.Instance)
+                            return Activator.CreateInstance(dstType, ToString(value));
+                        return null;
+                    }
 
                     if (dstType.GetTypeInfo().IsGenericType && dstType.Is(typeof(Nullable<>)))
                         return DecodeIntern(value, Nullable.GetUnderlyingType(dstType));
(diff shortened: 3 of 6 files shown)


kcrandall:
I've only done the Encoder part in the PR so far. Need guidance as to how the best way to detect the Fauna string in the decoder and then call the string constructor of the object, Uri or Guid, for example.

kcrandall:
Added tests that are passing and added the decoder functionality as well with a test

FaunaDB.Client/Types/Decoder.cs
```diff
@@ -93,6 +93,14 @@ static object DecodeIntern(Value value, Type dstType)
                 throw invalidCast(value, dstType);
             }
 
+            if (forceType == typeof(StringV) && value.GetType() == typeof(StringV))
+            {
+                if (value != null && value != NullV.Instance)
+                    return Activator.CreateInstance(dstType, ToString(value));
```
marrony, line 99: Really nice

FaunaDB.Client.Test/EncoderTest.cs
```diff
@@ -352,5 +352,33 @@ public void TestDateTimeOverride()
                 Encode(new DateTimeOverride(testDateTime, testDateTime))
             );
         }
+
+        class StringOverride
+        {
+            [FaunaField("uri")]
+            [FaunaString]
```
marrony, line 359: Can you add other tests ensuring it works on parameters as well (either on constructor or static method)

kcrandall: I'm not sure exactly what you mean. Can you give me an example?

marrony: https://github.com/fauna/faunadb-csharp/blob/master/FaunaDB.Client.Test/DecoderTest.cs#L194
https://github.com/fauna/faunadb-csharp/blob/master/FaunaDB.Client.Test/DecoderTest.cs#L339

kcrandall: Added a test and made it work with that

kcrandall: also upgraded it to work with primitive types and not just objs with string constructors

FaunaDB.Client/Types/Decoder.cs
```diff
@@ -93,6 +95,72 @@ static object DecodeIntern(Value value, Type dstType)
...
+                            if (typeof(DateTimeOffset) == dstType)
+                                return new DateTimeOffset(Convert.ToDateTime(ToString(value)));
+
+                            else if (typeof(byte[]) == dstType && value is BytesV)
+                                return JsonConvert.DeserializeObject<byte[]>(ToString(value));
+
+                            else if (dstType.IsArray && dstType.HasElementType)
+                                return JsonConvert.DeserializeObject(ToString(value), dstType);
+
+                            else if (typeof(IDictionary).IsAssignableFrom(dstType) || (dstType.GetTypeInfo().IsGenericType && dstType.Is(typeof(IDictionary<,>))))
```
marrony, line 113: This switch/case is a repetition, also I have the impression you can inline the deserialization in the block that already exists below it, like:
```csharp
                    if (typeof(IDictionary).IsAssignableFrom(dstType) || (dstType.GetTypeInfo().IsGenericType && dstType.Is(typeof(IDictionary<,>))))
                    {
                        if (forceType == typeof(StringV) && value.GetType() == typeof(StringV))
                            return JsonConvert.DeserializeObject(ToString(value), dstType);

                        return FromDictionary(value, dstType);
                    }
```
and, given you are adding that "serialization" feature, I would like to see tests covering it (encoding and decoding).

Another thing, what would be the use case for this serialization/deserialization? I liked the part where you call a string constructor for objects like `Url` and `Guid`, but I'm not sure I like this json serialization/deserialization.

kcrandall: How else would one represent collection like data as a string? Fauna String should work on most types if people want to store it as a string. If never store a collection like that, but somebody else might have a reason to put fauna string on a collection.

kcrandall: How else would one represent collection like data as a string? Fauna String should work on most types if people want to store it as a string. If never store a collection like that, but somebody else might have a reason to put fauna string on a collection.

kcrandall: Also I can take out the switch and put it all in the other block but there will be a lot of redundancies there too. You can't get around that here really. Like in every case there will be another if statement: if (forceType == typeof(StringV) && value.GetType() == typeof(StringV))

marrony: actually not every case, just for `TypeCode.Object`, case like `TypeCode.Single`, `TypeCode.Double` and etc are handled automatically because `ToNumber` function return strings as well and the `Convert` functions accept strings

kcrandall: I deleted the switch and moved it inline

FaunaDB.Client/Types/Decoder.cs
```diff
@@ -143,22 +145,66 @@ static object DecodeIntern(Value value, Type dstType)
...
                 case TypeCode.Object:
                     if (typeof(DateTimeOffset) == dstType)
-                        return new DateTimeOffset(ToDateTime(value));
+                        if (forceType == typeof(StringV) && value.GetType() == typeof(StringV))
+                            if (value != null && value != NullV.Instance)
+                                return new DateTimeOffset(Convert.ToDateTime(ToString(value)));
+                            else
+                                return null;
+                        else
+                            return new DateTimeOffset(ToDateTime(value));
```
marrony, line 154:
```suggestion
                    if (typeof(DateTimeOffset) == dstType)
                    {
                         if (forceType == typeof(StringV) && value.GetType() == typeof(StringV))
                         {
                             if (value != null && value != NullV.Instance)
                                 return new DateTimeOffset(Convert.ToDateTime(ToString(value)));
                             return null;
                         }
                         return new DateTimeOffset(ToDateTime(value));
                    }
```

FaunaDB.Client/Types/Decoder.cs
```diff
@@ -143,22 +145,66 @@ static object DecodeIntern(Value value, Type dstType)
 
                 case TypeCode.Object:
                     if (typeof(DateTimeOffset) == dstType)
-                        return new DateTimeOffset(ToDateTime(value));
+                        if (forceType == typeof(StringV) && value.GetType() == typeof(StringV))
+                            if (value != null && value != NullV.Instance)
+                                return new DateTimeOffset(Convert.ToDateTime(ToString(value)));
+                            else
+                                return null;
```
marrony, line 152: returning `null` for a primitive type will throw an error, I think you should return the default value for the type

FaunaDB.Client/Types/Decoder.cs
```diff
@@ -143,22 +145,66 @@ static object DecodeIntern(Value value, Type dstType)
...
+                        else
+                            return new DateTimeOffset(ToDateTime(value));
 
                     if (typeof(byte[]) == dstType && value is BytesV)
-                        return ToByteArray(value);
+                        if (forceType == typeof(StringV) && value.GetType() == typeof(StringV))
+                            if (value != null && value != NullV.Instance)
+                                return JsonConvert.DeserializeObject<byte[]>(ToString(value));
```
marrony, line 159: I'm still not convinced this is the correct way of doing this, in my opinion users could accomplish the same thing doing:
```csharp
        class Address
        { }

        class User
        {
            [FaunaField("addresses")]
            public string AddressesStr
            {
                get
                {
                    return JsonConvert.SerializeObject(Addresses);
                }

                set
                {
                    Addresses = JsonConvert.DeserializeObject<List<Address>>(value);
                }
            }

            [FaunaIgnore]
            public List<Address> Addresses { get; set; }
        }
```
this is more verbose but it will keep the driver simpler and give more flexibility to the user.

kcrandall: My view is the whole point of a driver is to make the code that uses it less verbose and much easier to use. Ive been doing that exact thing to get around for Uri types. I mean for me the Collections are a non issue but someone in the future might find it useful. Your previous suggestions have been resolved in 052582c

FaunaDB.Client/Types/Attributes.cs
```diff
@@ -92,7 +92,11 @@ public class FaunaDate : Attribute
     /// </example>
     [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property | AttributeTargets.Parameter)]
     public class FaunaTime : Attribute
-    {}
+    { }
+
```
evbruno, line 96: please add the proper doc/example here, such as `FaunaTime`

FaunaDB.Client/Types/Encoder.cs
```diff
@@ -85,6 +87,24 @@ Value EncodeIntern(object obj, Type forceType)
...
+                        if (typeof(byte[]).IsInstanceOfType(obj))
+                            return JsonConvert.SerializeObject(obj);
+
+                        if (typeof(IDictionary).IsInstanceOfType(obj))
+                            return JsonConvert.SerializeObject(obj);
+
+                        if (typeof(IEnumerable).IsInstanceOfType(obj))
+                            return JsonConvert.SerializeObject(obj);
+                        break;
+                }
```
evbruno, line 104: please add a test for this specific "branch"

FaunaDB.Client/Types/Decoder.cs
evbruno: Ok. In a second look, I'm in favor of Marrony's comment as well. Maybe we could remove that specific serialisation use cases from the Encoder/Decoder and add some sort of utility helper for the user. 

Lets try to figure it out a simpler "interface" to enable users to (de)serialize those types.

kcrandall: removed that

FaunaDB.Client/Types/Encoder.cs
kcrandall: removed no longer needs test

evbruno:
> @kcrandall please, rebase your code to master, we tried to do this locally and found a lot of conflicts

could you also "squash" the commits in a single one? I will  merge after that

kcrandall:
I dont usually do this much complex git stuff. Did that rebase work? just force pushed

kcrandall:
I squashed my commits

FaunaDB.Client/Types/Attributes.cs
```diff
@@ -92,7 +92,55 @@ public class FaunaDate : Attribute
...
+    ///     [FaunaString]
+    ///     public Uri MyLink { get; set; }
+    /// }
+    ///
+    /// var web = new Website { MyLink = new Uri("http://fauanadb.com") };
+    ///
+    /// var encodedWebsite = Encoder.Encode(web);
+    /// </example>
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property | AttributeTargets.Parameter)]
+    public class FaunaString : Attribute
```
marrony, line 142: You duplicated that class on the rebase

marrony:
Looks like you haven’t squashed your commits, you duplicated it, you have now 16 commits

FaunaDB.Client/Types/Attributes.cs
kcrandall: sorry something weird was going on i had to revert back and then rebased again and squashed

kcrandall:
Friendly reminder to close the issue I opened about fauna string too. Thanks for merging

PR #101, opened 2020-04-30
[DRV-129] add httpClient task timeout

evbruno:


FaunaDB.Client.Test/TimeoutClientTest.cs
```diff
@@ -0,0 +1,116 @@
...
+            Assert.ThrowsAsync<TimeoutException>(async () => await task0);
+
+            watch.Stop();
+
+            Assert.IsTrue(task0.IsCompleted, "the task0 is completed");
+            Assert.IsTrue(task0.IsFaulted, "the task0 failed");
+
+            StringAssert.Contains("The operation has timed out.", task0.Exception.Message);
+
+            Assert.IsTrue(watch.Elapsed.TotalSeconds >= 10 && watch.Elapsed.TotalSeconds <= 12);
```
evbruno, line 54: I'm not sure about that assertion. Maybe its overkill at this point, but the older impl. will take ~60s to cancel the task, throwing a `TaskCanceledException` instead of `TimeoutException`

PR #102, opened 2020-05-05
Update README

evbruno:


README.md
```diff
@@ -112,6 +112,16 @@ var client = new FaunaClient(endpoint: ENDPOINT, secret: SECRET, httpClient: HTT
 
 Except `secret` all other arguments are optional.
 
+You can also pass a custom HttpClient when creating a new FaunaClient:
+
+```csharp
+var http = new HttpClient();
+http.DefaultRequestHeaders.Add("User-Agent", "C# console program");
```
nathaniel-may, line 119: Are these exact string params required? Or are they just examples. We should clarify what we require / expect.

evbruno: This string is an example of a header that can be passed with a custom client. I tried to exemplify this by saying `You can also pass a custom HttpClient when creating a new FaunaClient`. I will appreciate any help with clarifying it :)

We could extend the example by adding a known header (such as `User-Agent`) and a custom one (such as `X-My-Awesome-Header`)

nathaniel-may: My instinct is to put a code comment in the example like 
```
// The default request headers can be any string values, but should be specific to your application.
```
If it's standard to use the `X-Word-Word` format the example should use that format too.

evbruno: Updated the example w/ a new comment about the header and the namespace for the client (to avoid any confusion)

PR #110, opened 2020-06-25
DRV-189 Refactor Contains into family of functions

evbruno:
notes:
- 1 test is expected to fail on `nightly`
- 2 tests are expected to fail on `stable`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ commands:
       - run:
           name: Wait FaunaDB init
           command: |
-            while ! $(curl --output /dev/null --silent --fail http://core:8443/ping); do sleep 1; done
+            while ! $(curl --output /dev/null --silent --fail -m 1 http://core:8443/ping); do sleep 1; done
 
       - run:
           name: Run Tests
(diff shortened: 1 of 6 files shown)


FaunaDB.Client/Query/Language.Miscellaneous.cs
```diff
@@ -176,25 +176,54 @@ public partial struct Language
         /// See the <see href="https://app.fauna.com/documentation/reference/queryapi#miscellaneous-functions">FaunaDB Miscellaneous Functions</see>
         /// </para>
         /// </summary>
+        [Obsolete("Contains is deprecated, please use ContainsField instead.")]
```
marrony, line 179: `ContaisPath`

.circleci/config.yml
```diff
@@ -32,7 +32,7 @@ commands:
       - run:
           name: Wait FaunaDB init
           command: |
-            while ! $(curl --output /dev/null --silent --fail http://core:8443/ping); do sleep 1; done
+            while ! $(curl --output /dev/null --silent --fail -m 1 http://core:8443/ping); do sleep 1; done
```
marrony, line 35: I think `--max-time 1` is more readable and compatible with previous arguments

PR #112, opened 2020-07-03
DRV-214 api_version on query values

evbruno:


FaunaDB.Client.Test/ClientTest.cs
```diff
@@ -1501,7 +1501,8 @@ public async Task TestEchoQuery()
             Assert.That(((QueryV)result).Value, Is.EquivalentTo(new Dictionary<string, Expr>
             {
                 {"lambda", Arr("x", "y")},
-                {"expr", Concat(Arr(Var("x"), "/", Var("y")))}
+                {"expr", Concat(Arr(Var("x"), "/", Var("y")))},
+                { "api_version", "3" }
```
lregnier, line 1505: This should be an integration test, right? :)

evbruno: This class is testing the "whole client", so it can be seen as a integration test (with lots of stuff on it). We definitely ~could~ need to improve those tests and refactor'em out in smaller classes.

PR #115, opened 2020-07-15
DRV-229: Allow implicit conversion of byte arrays

marrony:


FaunaDB.Client/Query/Expr.Operators.cs
```diff
@@ -33,6 +33,9 @@ public abstract partial class Expr
         public static implicit operator Expr(Dictionary<string, Expr> dict) =>
             Encoder.Encode(dict);
 
+        public static implicit operator Expr(byte[] bytes) =>
+            BytesV.Of(bytes);
+
```
evbruno, line 38: leet...

marrony:
I added a new function to the query api to wrap bytes type too
```
Value result = await client.Query(Bytes(0x1, 0x2, 0x3));
```

PR #116, opened 2020-07-15
DRV-228 update Fauna types ToString

evbruno:


FaunaDB.Client.Test/EncoderTest.cs
```diff
@@ -397,5 +397,31 @@ public void TestStringOverride()
...
+        {
+            Assert.AreEqual(
+                "SetRefV({match: RefV(id = \"spells_by_element\", collection = RefV(id = \"indexes\")), terms: StringV(water)})",
+                new SetRefV(new Dictionary<string, Value>() {
+                    { "terms", StringV.Of("water") },
+                    { "match", new RefV("spells_by_element", new RefV("indexes")) }
+                }).ToString()
+            );
+
+            Assert.AreEqual("BooleanV(True)", BooleanV.Of(true).ToString());
```
marrony, line 412: my vote is to remove the class name from the string representation, @faunaee will this break your tests?

faunaee: It won't break them, but I will have to update the expected output accordingly. I'm having to do that a fair bit anyway, so it's no problem. The opaque result was the problem.

evbruno: update: 
- simplified output (unwrapped the values)
- added `[]` to the Array
- dropped the `V` suffix for the remaining types (`Byte` and `Null`)

FaunaDB.Client.Test/EncoderTest.cs
```diff
@@ -397,5 +397,31 @@ public void TestStringOverride()
...
+                }).ToString()
+            );
+
+            Assert.AreEqual("true", BooleanV.Of(true).ToString());
+            Assert.AreEqual("3.14", DoubleV.Of(3.14).ToString());
+            Assert.AreEqual("42", LongV.Of(42).ToString());
+            Assert.AreEqual("null", NullV.Instance.ToString());
+            Assert.AreEqual("Date(\"2001-01-01\")", new DateV("2001-01-01").ToString());
+            Assert.AreEqual("Time(\"2000-01-01T01:10:30.123Z\")", new TimeV("2000-01-01T01:10:30.123Z").ToString());
+            Assert.AreEqual("Arr(1, 2, 3)", ArrayV.Of(1, 2, 3).ToString());
```
marrony, line 418: I think using brackets will be more common, `[1, 2, 3]`

evbruno: updated the `ArrayV`

evbruno:
rebased and squashed

faunaee:
During a fresh run of the example output, I can see two areas for improvement:

## Strings

Old:
```
StringV(munchings)
```

New:
```
munchings
```

Much nicer! However, within an array:

Old:
```
Arr(StringV(1234.5678), StringV(true), StringV(2018-06-06), StringV(null), StringV(2015-02-20T06:30:00Z))
```

New:
```
[1234.5678, true, 2018-06-06, null, 2015-02-20T06:30:00Z]
```

It is now impossible to tell, visually, whether that's the string "true" or the boolean value `true`.

For objects:

Old:
```
ObjectV(a: StringV(Apple),b: StringV(Banana),x: StringV(width),y: StringV(height))
```

New:
```
{"a": Apple, "b": Banana, "x": width, "y": height}
```

The string values now look like variables.

I'd suggest that within objects/arrays, strings should be quoted.

## QueryV

Old:
```
QueryV(System.Collections.Generic.Dictionary`2[System.String,FaunaDB.Query.Expr])
```

New:
```
System.Collections.Generic.Dictionary`2[System.String,FaunaDB.Query.Expr]
```

Neither representation is conveying anything useful to the developer.

faunaee:
Most everything looks good now. Since we are affecting most example output with this change, there is one more adjustment that could be made. Could we change the output for `Ref` values from:

```
Ref(id = "1", collection = Ref(id = "tokens"))
```

to:

```
Ref(id="1", collection=Ref(id="tokens"))
```

The difference is just the space surrounding the equals signs. It's a nice-to-have, not a need-to-have, but would provide more consistency in the whitespace included in query output that contains one or more `Ref`s.

faunaee:
I think that string handling needs some additional improvement:

In `fauna-shell`:
```
Create(Ref(Collection("test"), "2"), { data: { name: "test", text: 'test "quotes"\nmore\nextra' }})
```

And in a C# client program:
```
Value result = await client.Query(
    Get(Ref(Collection("test"), "2"))
);
Console.WriteLine(result);
```
the output is:
```
{"data": {"name": "test", "text": "test "quotes"
more
extra"}, "ref": Ref(id = "2", collection = Ref(id = "test", collection = Ref(id = "collections"))), "ts": 1596211054250000}
```

The quotations inside the string field likely need to be escaped.

evbruno:
escaping quotes for string done 👍

faunaee:
Looks good to me! I'm currently updating DOCS-1323 to match the latest changes, but I think it's now good to go.

evbruno:
rebased and squashed all the commits

PR #131, opened 2021-02-23
DRV-490: Streaming

parkhomenko:


parkhomenko:
Created another one in original repo:
https://github.com/fauna/faunadb-csharp/pull/132

PR #135, opened 2021-03-23
DRV-542: Add runtime environment/partner information to headers

parkhomenko:
### Notes
Collects runtime environment information and places it into "X-Driver-Env" header with every fauna request.
The information we collect is:
- operating system
- .net runtime version
- version of fauna driver
- specific environment information (aws lambda, vercel, google cloud, etc.)

There is also some clean-up for target frameworks in FaunaDB.Client and FaunaDB.Client.Test projects:
FuanaDB.Client is now targetting:
- net45 (this library can then be used for all .net framework 4.x versions since we don't use any specific API above .net framework 4.5 version and .net docs mention that all versions up to 4.8 are backward compatible with 4.5)
- netstandard1.5 (will be used with .net core 1.0, 1.1, and others applicable to this standard)
- netstandard2.0 (will be used with .net core 2.0, 2.1, 2.2, and others applicable to this standard)
- netstandard2.1 (will be used with .net core 3.0, 3.1, .net5.0, and others applicable to this standard)
FaunaDB.Client.Test is now targetting:
- net45 (no need for >net45 since they are backward compatible and we don't use any specific API above .net framework 4.5)
- .net core 2.1
- .net core 3.1
- .net 5.0
- we'll need to add .net 6.0 (when it is released since it is the next LTS version)

### Jira ticket
https://faunadb.atlassian.net/browse/DRV-542

### DIFF
--- a/FaunaDB.Client/Client/DefaultClientIO.cs
+++ b/FaunaDB.Client/Client/DefaultClientIO.cs
@@ -59,7 +59,7 @@ namespace FaunaDB.Client
             message.Headers.Authorization = authHeader;
             message.Headers.AcceptEncoding.Add(new StringWithQualityHeaderValue("gzip"));
             message.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
-            message.Headers.Add("X-FaunaDB-API-Version", "3");
+            message.Headers.Add("X-FaunaDB-API-Version", "4");
             message.Headers.Add("X-Fauna-Driver", "csharp");
 
             var last = lastSeen.Txn;
@@ -112,12 +112,11 @@ namespace FaunaDB.Client
             headers.ToDictionary(k => k.Key, v => v.Value);
 
         /// <summary>
-        /// Encodes secret string using base64.
+        /// Adds a header with Bearer Auth Token.
         /// </summary>
-        static AuthenticationHeaderValue AuthHeader(string secret)
+        static AuthenticationHeaderValue AuthHeader(string authToken)
         {
-            var bytes = Encoding.ASCII.GetBytes(secret);
-            return new AuthenticationHeaderValue("Basic", Convert.ToBase64String(bytes));
+            return new AuthenticationHeaderValue("Bearer", authToken);
         }
 
         /// <summary>
--- a/FaunaDB.Client/Client/DefaultClientIO.cs
+++ b/FaunaDB.Client/Client/DefaultClientIO.cs
@@ -67,6 +67,7 @@ namespace FaunaDB.Client
             message.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
             message.Headers.Add("X-FaunaDB-API-Version", "4");
             message.Headers.Add("X-Fauna-Driver", "csharp");
+            message.Headers.Add("X-Driver-Env", RuntimeEnvironmentHeader.Construct());
 
             var last = lastSeen.Txn;
             if (last.HasValue)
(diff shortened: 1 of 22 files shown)


FaunaDB.Client/Client/RuntimeEnvironmentHeader.cs
```diff
@@ -0,0 +1,167 @@
...
+        private static RuntimeEnvironmentHeader instance;
+
+        private RuntimeEnvironmentHeader() {}
+
+        private void GatherEnvironmentInfo(IEnvironmentEditor environmentEditor)
+        {
+            this.environment = GetRuntimeEnvironment(environmentEditor);
+            this.operatingSystem = GetOperatingSystemName();
+            this.runtime = GetCurrentRuntime();
+            this.driverVersion = Assembly.Load(new AssemblyName("FaunaDB.Client")).GetName().Version.ToString();
```
marrony, line 23: can you use `Assembly.GetExecutingAssembly()`?

FaunaDB.Client/Client/RuntimeEnvironmentHeader.cs
```diff
@@ -0,0 +1,167 @@
...
+            {
+                return "Linux";
+            }
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                return "Windows";
+            }
+            return "Unknown";
+#else
+            // if we're under one of the .net frameworks than OS is Windows
```
marrony, line 43: I don't think I understood this affirmation, what happens if I run on macosx using mono with net45 framework? If I understood the code it will detect as Windows?

FaunaDB.Client/Client/DefaultClientIO.cs
```diff
@@ -117,6 +120,7 @@ async Task<StreamingRequestResult> DoStreamingRequestAsync(string data, IReadOnl
             message.Headers.Authorization = authHeader;
             message.Headers.Add("X-FaunaDB-API-Version", "4");
             message.Headers.Add("X-Fauna-Driver", "csharp");
```
n400, line 122: Aren't we removing `X-Fauna-Driver` since "c-sharp-*" is now being sent in `X-Runtime-Env`?

FaunaDB.Client/Client/RuntimeEnvironmentHeader.cs
parkhomenko: I tried to use this function at first, but it doesn't work if I build the project against `dotnetstandard1.5`:
```
error CS0117: 'Assembly' does not contain a definition for 'GetExecutingAssembly'
```
This is happening because there is a conditional library, which I think overrides the standard `System.Reflections` one:
https://github.com/fauna/faunadb-csharp/blob/master/FaunaDB.Client/FaunaDB.Client.csproj#L32

I even tried to get rid of the dotnetstandard1.5, but it is specifically used here:
https://github.com/fauna/faunadb-csharp/blob/master/FaunaDB.Client/Types/Encoder.cs#L80-L85
and two more places. It looks like some workaround. I can be mistaken, but it looks like we can create another ticket to avoid overriding `System.Reflection` for dotnetstandard1.5.

FaunaDB.Client/Client/RuntimeEnvironmentHeader.cs
parkhomenko: I was thinking about it as well, but the function `Environment.OSVersion.ToString();` didn't work for me at first.
For some reason, it works now, probably after target frameworks clean-up.
Thanks for pointing this out, I've changed the code so it can detect environments beside Windows

FaunaDB.Client/Client/DefaultClientIO.cs
parkhomenko: thanks for this one, I've removed the `X-Fauna-Driver` header

evbruno (approved):

PR #137, opened 2021-03-31
DRV-47: Update C# driver to speak HTTP/2

parkhomenko:
### Notes
Adds an optional parameter to FaunaClient, so user can support http version:
```
var client = new FaunaClient(endpoint: ENDPOINT, secret: SECRET, httpVersion: HttpVersion.Version20);
```

### Jira ticket
https://faunadb.atlassian.net/browse/DRV-47

### DIFF
--- a/FaunaDB.Client/Client/DefaultClientIO.cs
+++ b/FaunaDB.Client/Client/DefaultClientIO.cs
@@ -39,7 +39,11 @@ namespace FaunaDB.Client
             this.lastSeen = lastSeen;
             this.endpoint = endpoint;
             this.clientTimeout = timeout;
+#if NETSTANDARD2_1
+            this.httpVersion = httpVersion == null ? new Version(2, 0) : httpVersion;
+#else
             this.httpVersion = httpVersion == null ? new Version(1, 1) : httpVersion;
+#endif
         }
 
         public DefaultClientIO(string secret, Uri endpoint, TimeSpan? timeout = null, HttpClient httpClient = null, Version httpVersion = null)
--- a/FaunaDB.Client/Client/DefaultClientIO.cs
+++ b/FaunaDB.Client/Client/DefaultClientIO.cs
@@ -39,7 +39,7 @@ namespace FaunaDB.Client
             this.lastSeen = lastSeen;
             this.endpoint = endpoint;
             this.clientTimeout = timeout;
-            this.httpVersion = httpVersion == null ? HttpVersion.Version11 : httpVersion;
+            this.httpVersion = httpVersion == null ? new Version(1, 1) : httpVersion;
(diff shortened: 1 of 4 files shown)


FaunaDB.Client/Client/DefaultClientIO.cs
```diff
@@ -27,25 +27,31 @@ class DefaultClientIO : IClientIO
...
             this.client = client;
             this.authHeader = authHeader;
             this.lastSeen = lastSeen;
             this.endpoint = endpoint;
             this.clientTimeout = timeout;
+#if NETSTANDARD2_1
+            this.httpVersion = httpVersion == null ? new Version(2, 0) : httpVersion;
+#else
+            this.httpVersion = httpVersion == null ? new Version(1, 1) : httpVersion;
+#endif
```
evbruno, line 46: oh, nice catch!

evbruno: btw... this is going to be valid for standards >= 2.1, right? 
(that framework/standard matrix is always confusing me)

README.md
```diff
@@ -146,6 +146,19 @@ http.Timeout = TimeSpan.FromSeconds(15);
...
+#### HTTP 2.0 support
+Starting from version 4.0.0 of this driver (faunadb-csharp), HTTP/2 support is enabled by default for .NET standards 2.1 and above.
+This means that if you use .NET core 3.1 and above (which support that standard), you'll be sending requests to Fauna on HTTP/2. 
+.NET standards lower than 2.1 and .NET frameworks 4.5-4.8 have HTTP/1.1 enabled as the default protocol version, since they lack of support for HTTP/2.
+We've also added an optional parameter if you want to specify the version of the protocol directly:
+```
+var adminClient = new FaunaClient(
+    endpoint: endpoint,
+    secret: secret,
+    httpVersion: HttpVersion.Version20
```
evbruno, line 158: as `2.0` is the current default, I think the example could be showing the `1.1`, wdyt?

FaunaDB.Client/Client/DefaultClientIO.cs
parkhomenko: yep, they added http/2 support only starting from .net core 3.0, which is standard 2.1,
the above standards will also support http/2 (there is not going to be standards notion anymore after 2.1 though :) they moved to .net 5, .net 6 and so on, but that can be considered as "above")

README.md
parkhomenko: that makes sense, thank you
I've changed to 1.1

evbruno (approved):
🚀

vadimLF (approved):
LGTM

PR #138, opened 2021-04-02
ROAD-502: Release 4.0.0-rc

parkhomenko:
Release candidate version for C# driver

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - Streaming support
 - 3rd party auth support
 - HTTP/2 support for .net standard 2.1 and above
+- Add functions: `All`, `Any`, `IsArray`, `IsBoolean`, `IsBytes`, `IsCollection`, `IsCredentials`, `IsDatabase`, `IsDate`, `IsDoc`, `IsDouble`, `IsFunction`, `IsIndex`, `IsInteger`, `IsKey`, `IsLambda`, `IsNull`, `IsNumber`, `IsObject`, `IsRef`, `IsRole`, `IsSet`, `IsString`, `IsTimestamp`, `IsToken`, `TimeAdd`, `TimeSubtract`, `TimeDiff`, `ToArray`, `ToDouble`, `ToInteger`, `ToObject`
 
 ## 3.0.0
 - Deprecate `Contains` in favor of specific functions `ContainsField`, `ContainsPath`, and `ContainsValue`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 4.0.0-rc
+- Streaming support
+- 3rd party auth support
+- HTTP/2 support for .net standard 2.1 and above
+
 ## 3.0.0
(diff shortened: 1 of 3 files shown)


faunaee:
Note that the C# driver is not appear to provide implementations for the following functions:

* All
* Any
* IsArray
* IsBoolean
* IsBytes
* IsCollection
* IsCredentials
* IsDatabase
* IsDate
* IsDoc
* IsDouble
* IsFunction
* IsIndex
* IsInteger
* IsKey
* IsLambda
* IsNull
* IsNumber
* IsObject
* IsRef
* IsRole
* IsSet
* IsString
* IsTimestamp
* IsToken
* TimeAdd
* ToArray
* ToDouble
* ToInteger
* ToObject

These functions were all added to FQL prior to API v3, so it would be a major omission if they were (still) not implemented in a v4 driver.

n400:
For Ewan’s note: https://faunadb.atlassian.net/browse/DRV-598

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 4.0.0-rc
+- Streaming support
+- 3rd party auth support
+- HTTP/2 support for .net standard 2.1 and above
```
n400, line 4: List all of the missing functions that got added?

n400 (changes requested):

CHANGELOG.md
parkhomenko: added

FaunaDB.Client/FaunaDB.Client.csproj
```diff
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>net45;netstandard1.5;netstandard2.0;netstandard2.1</TargetFrameworks>
-    <Version>3.0.0</Version>
+    <Version>4.0.0-rc</Version>
     <PackageId>FaunaDB.Client</PackageId>
```
evbruno, line 6: dumb question here: is the `rc` the same purpose of "snapshots" for Maven repo (JVM) ? Can we override `rc` releases?

evbruno (approved):

FaunaDB.Client/FaunaDB.Client.csproj
parkhomenko: I think they have different meanings. Maven's snapshot version is not considered a release version and can be updated, so a user who uses 3.0.0-snapshot, for example, can expect that there can be new changes within this same version.
Nuget's `rc` can't be changed in the same way, if we have new changes we need to publish another package, with, say, version 4.0.0-rc1, 4.0.0-rc2, and so on.

We can write anything after the "-", but conventionally we use alpha, beta, rc if we mean pre-release version. There should be a checkbox in visual studio to include pre-release version, so users can use them in their projects.
And they are also marked as pre-release on nuget.org:
https://www.nuget.org/packages/FaunaDB.Client/1.2.1-alpha

n400 (approved):

updatus (approved):
LGTM.

vadimLF (approved):
LGTM.

PR #149, opened 2021-05-14
OSS-694: Validate input parameters

shestakovg:
…thods

### DIFF
--- a/FaunaDB.Client/Client/DefaultClientIO.cs
+++ b/FaunaDB.Client/Client/DefaultClientIO.cs
@@ -9,11 +9,7 @@ using System.Net.Http.Headers;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
-<<<<<<< HEAD
-using FaunaDB.Collections;
 using FaunaDB.Errors;
-=======
->>>>>>> master
 
 namespace FaunaDB.Client
 {
--- a/FaunaDB.Client/Client/DefaultClientIO.cs
+++ b/FaunaDB.Client/Client/DefaultClientIO.cs
@@ -34,6 +34,22 @@ namespace FaunaDB.Client
 
         internal DefaultClientIO(HttpClient client, AuthenticationHeaderValue authHeader, LastSeen lastSeen, Uri endpoint, TimeSpan? timeout, Version httpVersion)
         {
+            if (client == null)
+            {
+                throw new ArgumentNullException(nameof(client));
+            }
+            if (authHeader == null)
+            {
+                throw new ArgumentNullException(nameof(authHeader));
+            }
+            if (endpoint == null)
+            {
+                throw new ArgumentNullException(nameof(endpoint));
+            }
+            if (lastSeen == null)
+            {
+                throw new ArgumentNullException(nameof(lastSeen));
+            }
             this.client = client;
             this.authHeader = authHeader;
             this.lastSeen = lastSeen;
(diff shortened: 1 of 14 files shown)


parkhomenko (dismissed):

FaunaDB.Client/Client/DefaultClientIO.cs
```diff
@@ -34,6 +34,22 @@ class DefaultClientIO : IClientIO
...
+                throw new ArgumentNullException(nameof(authHeader));
+            }
+            if (endpoint == null)
+            {
+                throw new ArgumentNullException(nameof(endpoint));
+            }
+            if (lastSeen == null)
+            {
+                throw new ArgumentNullException(nameof(lastSeen));
+            }
```
evbruno, line 52: nice sanitizing !!

evbruno, line 52: can we create an auxiliary method so we can "compact" those calls, something like:
```          
            AssertNotNull( client );
            AssertNotNull( authHeader );
            AssertNotNull( endpoint );
            AssertNotNull( lastSeen );
```

FaunaDB.Client/Client/FaunaClient.cs
```diff
@@ -46,6 +46,10 @@ public class FaunaClient
         /// </summary>
         public FaunaClient(IClientIO clientIO)
         {
+            if (clientIO == null)
+            {
+                throw new ArgumentNullException(nameof(clientIO));
+            }
             this.clientIO = clientIO;
```
evbruno, line 53: same here

evbruno (changes requested):
just a minor comment; its not blocking the PR and you can dismiss if you like too

FaunaDB.Client/Client/DefaultClientIO.cs
shestakovg: Really it's good idea. Will do

FaunaDB.Client/Errors/NotNullValidator.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace FaunaDB.Errors
+{
+    static class NotNullValidator
+    {
+        public static void AssertNotNull(this object value, string valueName)
+        {
+            if (value == null)
+            {
+                throw new ArgumentNullException(valueName);
```
evbruno, line 11: can we use `nameof` here so the callers are even more simple?

```suggestion
                throw new ArgumentNullException(nameof(valueName));
```

evbruno (changes requested):
I think it would be nice to add some tests (at least for `FaunaClient`)

evbruno:
I'm still wondering if this is better accomplished w/ the extension method proposed (as there are no base types, only `object`) or a class method (static). Any reasons for that? (not a c# expert, though)

evbruno:
miss clicked on`Close` instead of `Comment`, my bad 🤦

evbruno (changes requested):
there are some comments/questions on the PR, but the main change that I would like to see is (at least) one test for the inputs; if so, the `FaunaClient` constructor should suffice 😉

evbruno (dismissed):
🚢 -it

parkhomenko (approved):

parkhomenko (approved):

vadimLF (approved):
LGTM. Just confusing with extension method name which looks like some xUnit method (maybe checkNotNull maybe ?)

PR #153, opened 2021-06-07
OSS-665 driver notification

shestakovg:
### Overview
Get latest NuGet version from server and compare it with local version. 
Added ability to turn of new version checking. Parameter dontCheckNewVersion in constructor.

### Jira ticket
https://faunadb.atlassian.net/browse/OSS-665

### DIFF
--- a/FaunaDB.Client/Client/DefaultClientIO.cs
+++ b/FaunaDB.Client/Client/DefaultClientIO.cs
@@ -47,7 +47,10 @@ namespace FaunaDB.Client
 #else
             this.httpVersion = httpVersion == null ? new Version(1, 1) : httpVersion;
 #endif
-            Task.Run(()=> FaunaDB.Client.Utils.CheckLatestVersion.GetVersionAsync()).Wait();
+            if (!Utils.CheckLatestVersion.AlreadyChecked.HasValue)
+            {
+                Task.Run(() => Utils.CheckLatestVersion.GetVersionAsync()).Wait();
+            }
         }
 
         public DefaultClientIO(string secret, Uri endpoint, TimeSpan? timeout = null, HttpClient httpClient = null, Version httpVersion = null)
--- a/FaunaDB.Client/Client/DefaultClientIO.cs
+++ b/FaunaDB.Client/Client/DefaultClientIO.cs
@@ -47,10 +47,6 @@ namespace FaunaDB.Client
 #else
             this.httpVersion = httpVersion == null ? new Version(1, 1) : httpVersion;
 #endif
-            if (!Utils.CheckLatestVersion.AlreadyChecked.HasValue)
-            {
-                Task.Run(() => Utils.CheckLatestVersion.GetVersionAsync()).Wait();
-            }
         }
 
         public DefaultClientIO(string secret, Uri endpoint, TimeSpan? timeout = null, HttpClient httpClient = null, Version httpVersion = null)
--- a/FaunaDB.Client/Client/DefaultClientIO.cs
+++ b/FaunaDB.Client/Client/DefaultClientIO.cs
@@ -47,6 +47,7 @@ namespace FaunaDB.Client
 #else
             this.httpVersion = httpVersion == null ? new Version(1, 1) : httpVersion;
 #endif
+            Task.Run(()=> FaunaDB.Client.Utils.CheckLatestVersion.GetVersionAsync()).Wait();
         }
 
         public DefaultClientIO(string secret, Uri endpoint, TimeSpan? timeout = null, HttpClient httpClient = null, Version httpVersion = null)
--- a/FaunaDB.Client/Client/FaunaClient.cs
+++ b/FaunaDB.Client/Client/FaunaClient.cs
@@ -8,6 +8,7 @@ using FaunaDB.Errors;
 using FaunaDB.Query;
 using FaunaDB.Types;
 using Newtonsoft.Json;
+using FaunaDB.Client.Utils;
 
 namespace FaunaDB.Client
 {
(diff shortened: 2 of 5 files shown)


FaunaDB.Client/Client/FaunaClient.cs
```diff
@@ -32,13 +33,15 @@ public class FaunaClient
...
+        /// <param name="dontCheckNewVersion">Don't check new NuGet package driver version. Default value is null (Check new version).</param>
         public FaunaClient(
             string secret,
             string endpoint = "https://db.fauna.com:443",
             TimeSpan? timeout = null,
             HttpClient httpClient = null,
-            Version httpVersion = null)
-            : this(CreateClient(secret, endpoint, timeout, httpClient, httpVersion))
+            Version httpVersion = null,
+            bool? dontCheckNewVersion = null)
```
parkhomenko, line 43: Can we have a default checkNewVersion parameter of `true` instead of negated one? In a similar way the `endpoint` parameter is made above.

FaunaDB.Client/Utils/CheckLatestVersion.cs
```diff
@@ -0,0 +1,70 @@
...
+            var url = $"https://api.nuget.org/v3-flatcontainer/{packageName}/index.json";
+            try
+            {
+                var httpClient = new HttpClient();
+#if NET45
+                ServicePointManager.SecurityProtocol = SecurityProtocolType.Ssl3 | SecurityProtocolType.Tls | SecurityProtocolType.Tls11 | SecurityProtocolType.Tls12;
+#endif
+                var response = await httpClient.GetAsync(url);
+                string versionsResponse = await response.Content.ReadAsStringAsync();
+                Newtonsoft.Json.Linq.JObject jObject = Newtonsoft.Json.Linq.JObject.Parse(versionsResponse);
```
parkhomenko, line 29: this is just a question:
why do we need to use the full name of the class, instead of placing it above to `using`?

FaunaDB.Client/Utils/CheckLatestVersion.cs
```diff
@@ -0,0 +1,70 @@
...
+                return;
+            }
+
+            Assembly asm = typeof(CheckLatestVersion).GetTypeInfo().Assembly;
+            var currentVersion =  asm.GetName().Version;
+            var currentVersionString = $"{currentVersion.Major}.{currentVersion.Minor}.{currentVersion.Build}";
+            if (!latestNuGetVesrionString.Equals(currentVersionString))
+            {
+                var message =  GetMessage(latestNuGetVesrionString, currentVersionString);
+                Debug.WriteLine(message);
```
parkhomenko, line 47: Is it a good practice to leave `Debug` messages in the production code in c#?

parkhomenko (changes requested):
Can we add some tests to check this functionality?
I would add at least 2 tests to check if we pass/don't pass a "checkNewVersion" param.

FaunaDB.Client/Utils/CheckLatestVersion.cs
shestakovg: Just didn't wont to use link to Newtonsoft. I remember we want to move to Text.Json.

FaunaDB.Client/Utils/CheckLatestVersion.cs
shestakovg: Sure. It works only in Debug mode. In release mode it doesn't work

FaunaDB.Client/Client/FaunaClient.cs
```diff
@@ -32,13 +33,15 @@ public class FaunaClient
         /// <param name="endpoint">URL for the FaunaDB server. Defaults to "https://db.fauna.com:443"</param>
         /// <param name="timeout">Timeout for I/O operations. Defaults to 1 minute.</param>
         /// <param name="httpVersion">Version of http. Default value is HttpVersion.Version11, is you use .net core 3.0 and above you can enable http/2 support by passing HttpVersion.Version20</param>
+        /// <param name="dontCheckNewVersion">Don't check new NuGet package driver version. Default value is null (Check new version).</param>
```
parkhomenko, line 36: You might also want to correct the comments and PR description for "checkNewVersion"

FaunaDB.Client/Utils/CheckLatestVersion.cs
```diff
@@ -0,0 +1,73 @@
...
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace FaunaDB.Client.Utils
+{
+    public class CheckLatestVersion
+    {
+        private const string packageName = "FaunaDB.Client";
+        public static bool? AlreadyChecked { get; set; } = null;
```
parkhomenko, line 15: Why we can't just set it to `false` as a default value instead of using a nullable type?

FaunaDB.Client/Client/FaunaClient.cs
```diff
@@ -245,10 +248,15 @@ public static ObjectV FromJson(string json)
             string endpoint,
             TimeSpan? timeout = null,
             HttpClient httpClient = null,
-            Version httpVersion = null)
+            Version httpVersion = null,
+            bool checkNewVersion = true)
         {
             secret.AssertNotNull(nameof(secret));
             endpoint.AssertNotNull(nameof(endpoint));
+            if (!checkNewVersion)
```
parkhomenko, line 256: This `if` doesn't look good. Can we just pass a `checkNewVersion` param into `DefaultClientIO` and add it into conditional there:
```
if (checkNewVersion && !Utils.CheckLatestVersion.AlreadyChecked.HasValue) 
```

FaunaDB.Client/Client/DefaultClientIO.cs
```diff
@@ -47,6 +47,10 @@ internal DefaultClientIO(HttpClient client, AuthenticationHeaderValue authHeader
 #else
             this.httpVersion = httpVersion == null ? new Version(1, 1) : httpVersion;
 #endif
+            if (!Utils.CheckLatestVersion.AlreadyChecked.HasValue)
+            {
+                Task.Run(() => Utils.CheckLatestVersion.GetVersionAsync()).Wait();
```
parkhomenko, line 52: Is `Wait` blocking the code at this point? Do we really need to block it here?

FaunaDB.Client/Client/FaunaClient.cs
shestakovg: Ooh, sure. My bad. I forgot to modify parameter name.

parkhomenko (approved):

vadimLF (approved):
The only concern is PR name - not sure that it's about load tests.

PR #155, opened 2021-07-01
object property mapping is now case insensitive

goncalo-oliveira:
This makes it possible to bind values to properties regardless of the casing, without the need to use the `FaunaField` attribute.

In sum.... this
```csharp
public class Store
{
    public string Name { get; set; }
}
```
instead of this
```csharp
public class Store
{
    [FaunaField( "name" ]
    public string Name { get; set; }
}
```


goncalo-oliveira:
I'm withdrawing this as it only works for decoding. When encoding, the field name will use the property casing. So, not a solution.

PR #158, opened 2021-07-21
feat(OSS-846): Added an optional customHeaders parameter for FaunaCli…

parkhomenko:
### Notes
We need to be able to set the X-Fauna-Source header in some applications that use FaunaClient for user statistics. This PR adds a customHeaders parameter, so users who use it can specify X-Fauna-Source header.

### Jira ticket
https://faunadb.atlassian.net/browse/OSS-846

### DIFF
--- a/FaunaDB.Client.Test/TestCase.cs
+++ b/FaunaDB.Client.Test/TestCase.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Net;
 using System.Net.Http;
 using System.Text;
@@ -21,8 +22,10 @@ namespace Test
         protected static Field<RefV> REF_FIELD = Field.At("ref").To<RefV>();
 
         protected FaunaClient rootClient;
+        protected FaunaClient rootClientWithCustomHeaders;
         protected Expr DbRef;
         protected FaunaClient client;
+        protected FaunaClient clientWithCustomHeaders;
         protected FaunaClient adminClient;
         protected Value clientKey;
         protected Value adminKey;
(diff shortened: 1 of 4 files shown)


FaunaDB.Client.Test/TestCase.cs
```diff
@@ -50,6 +53,15 @@ private async Task SetUpAsync()
...
             rootClient = new FaunaClient(secret: faunaSecret, endpoint: faunaEndpoint);
+            rootClientWithCustomHeaders = new FaunaClient(
+                secret: faunaSecret,
+                endpoint: faunaEndpoint,
+                customHeaders: new Dictionary<string, string>
+                {
+                    { "test-header-1", "test-value-1" },
+                    { "test-header-2", "test-value-2" },
+                }
+            );
```
evbruno, line 65: what do you think on moving the client creation only to the test itself? 
I don't think is necessary to have them as  members on the base class if we are only going to use once, to test the headers

evbruno: do we need both `rootClientWithCustomHeaders` and `clientWithCustomHeaders` ? I didn't find any use for `rootClientWithCustomHeaders`

evbruno (changes requested):
see comments above

FaunaDB.Client/Client/DefaultClientIO.cs
```diff
@@ -24,13 +24,15 @@ internal class DefaultClientIO : IClientIO
...
+
         private LastSeen lastSeen;
         private Version httpVersion;
 
         public const string StreamingPath = "stream";
         public const HttpMethodKind StreamingHttpMethod = HttpMethodKind.Post;
 
-        internal DefaultClientIO(HttpClient client, AuthenticationHeaderValue authHeader, LastSeen lastSeen, Uri endpoint, TimeSpan? timeout, Version httpVersion)
+        internal DefaultClientIO(HttpClient client, AuthenticationHeaderValue authHeader, LastSeen lastSeen, Uri endpoint, TimeSpan? timeout, Version httpVersion, IReadOnlyDictionary<string, string> customHeaders)
         {
```
evbruno, line 36: I don't think is necessary for the current PR, but WDYT about adding some builders to replace those methods/constructors w/ big arguments list? We have 7 here

parkhomenko: Yes, you're right. We actually use a builder in Java driver, it makes sense to it here as well.

FaunaDB.Client.Test/TestCase.cs
parkhomenko: makes sense, moved it to the test

evbruno (approved):

n400 (approved):

PR #167, opened 2021-08-30
OSS-890: Driver needs to detect and throw an error upon loss of networking

shestakovg:
### Overview
check connection state during a streaming and throw an error
### Jira ticket
https://faunadb.atlassian.net/browse/OSS-890

### DIFF
--- a/FaunaDB.Client/Client/StreamingEventHandler.cs
+++ b/FaunaDB.Client/Client/StreamingEventHandler.cs
@@ -2,6 +2,8 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 using FaunaDB.Errors;
 using FaunaDB.Types;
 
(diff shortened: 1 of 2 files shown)


FaunaDB.Client/Client/StreamingEventHandler.cs
```diff
@@ -2,25 +2,61 @@
...
+                            {
+                                observer.OnError(ex);
+                            }
+                        }
+
+                        Task.Delay(TIME_OUT, token).Wait();
+                    }
+                }, TaskCreationOptions.AttachedToParent
+                );
+            }
```
evbruno, line 59: is this formatting right ?

FaunaDB.Client/Client/StreamingEventHandler.cs
```diff
@@ -2,25 +2,61 @@
...
+using System.Threading;
+using System.Threading.Tasks;
 using FaunaDB.Errors;
 using FaunaDB.Types;
 
 namespace FaunaDB.Client
 {
     public class StreamingEventHandler : IObservable<Value>, IDisposable
     {
+        private const int TIME_OUT = 10000;
```
evbruno, line 14:
```suggestion
        private const int TIME_OUT_IN_MILLIS = 10000;
```
or something like that :D

evbruno (changes requested):
minor

evbruno:
is there an easy way to test this without mocking ?

FaunaDB.Client/Client/StreamingEventHandler.cs
shestakovg: Task should be completed before main task. But actually we doesn't need that (It works good without this param)

PR #175, opened 2021-10-29
Bumped to version 5.0.0

parkhomenko:


faunaee (changes requested):
The README.md does not describe the driver exception hierarchy or what it means for developers.

README.md
```diff
@@ -283,6 +283,42 @@ There are three attributes that can be used to change the behavior of the `Encod
 - Primitive arrays, generic collections such as `List<T>`, and their respective interfaces such as `IList<T>`.
 - Dictionaries with string keys, such as `Dictionary<string, T>` and its respective interface `IDictionary<string, T>`.
 
+### Exceptions handling
+
+Since 5.0.0 release, we introduced children classes for base FaunaException class,
+for example:
```
faunaee, line 289:
```suggestion
With the 5.0.0 release, child classes of the base FaunaException class now exist.
For example:
```

README.md
```diff
@@ -283,6 +283,42 @@ There are three attributes that can be used to change the behavior of the `Encod
...
+{
+    public InvalidArgumentException(int httpStatusCode, string message, string[] position) : base(httpStatusCode, ExceptionCodes.InvalidArgument, message, position) { }
+}
+
+public class TransactionAbortedException : FaunaException
+{
+    public TransactionAbortedException(int httpStatusCode, string message, string[] position) : base(httpStatusCode, ExceptionCodes.TransactionAborted, message, position) { }
+}
+```
+where each class corresponds to the error code from fauna database server,
```
faunaee, line 306:
```suggestion
Each class corresponds to an error code from Fauna.
```

README.md
```diff
@@ -283,6 +283,42 @@ There are three attributes that can be used to change the behavior of the `Encod
...
+you can inspect "FaunaException.cs" file for more information on how it is implemented.
```
faunaee, line 307:
```suggestion
Inspect the [`FaunaException.cs`](https://github.com/fauna/faunadb-csharp/blob/v5/FaunaDB.Client/Errors/FaunaException.cs)
file for more information on how it is implemented.
```

README.md
```diff
@@ -283,6 +283,42 @@ There are three attributes that can be used to change the behavior of the `Encod
...
+
+Below you can see an example of how you can take advantage of a new exception hierarchy:
```
faunaee, line 309:
```suggestion
The following example demonstrates the methods that you can access from the exception object:
```

README.md
```diff
@@ -283,6 +283,42 @@ There are three attributes that can be used to change the behavior of the `Encod
...
+```
+try
+{
+    adminClient.Query(ToDouble(Now()));
+}
+catch (InvalidArgumentException e)
+{
+    Console.WriteLine(e.HttpStatusCode);
```
faunaee, line 317: Should that be `httpStatusCode`?

README.md
```diff
@@ -283,6 +283,42 @@ There are three attributes that can be used to change the behavior of the `Encod
...
+    Console.WriteLine(e.Message);
```
faunaee, line 318: Is this the same as the Python driver's `e.description`? If so, can we rename it or add `description` as an alias?

What about the `e.errors` call from the Python driver? Is there an equivalent in C#? If not, can we add one?

faunaee (changes requested):
A few suggestions for language improvement, and a couple of questions about exception object attributes/methods.

README.md
parkhomenko: c# has a language feature, called attributes, that doesn't exist in java or python, and they write them starting with a capital letter, I think that will be confusing in .net world to have it written from a lowercase letter.

README.md
parkhomenko: not sure what we can do with the `description` field... Java and C# has a `message` field that is standard for any exception in language's SDK, I mean it is already defined, we just use it. We can have a duplicate `description` field, but that can look weird again for Java or C# devs

`e.errors` is an internal thing, I made a mistake showing it in the readme file for python, already removed it from there

README.md
faunaee: 👍

README.md
faunaee: How would a developer access the `position` and `cause` fields in error responses? AFAIK, `cause` only appears in the result when a called UDF has an error during its processing.

parkhomenko: You're right that the `cause` field is related to UDF (FunctionCallError). This is a special case. Another similar one is `ValidationFailed` error, where we have a field `failures`. All 2 fields are hidden from users.
For C# they get an array of `ValidationFailure` objects with `Code`, `Description` and `Position` fields
In Java we still get an array of `FaunaException` objects with `Code`, `Description` and Position` fields
a bit different implementation, but the general idea is the same

Regarding the `position` field: it can be accessed in the same way as `HttpStatusCode` and `Message` but when we discussed it with the Core team, they said that it is an experimental feature and can be removed in the future, so I thought it is ok to have the possibility to access it for users, but not include it in the readme file or as an official thing anywhere. That's just my opinion anyway, if you want I can add it.

faunaee: The goal of the technical specification (https://docs.google.com/document/d/1VKRmSfm6gqWjVwImORz8DiGRvitRTbvIckIwcnS57TA/edit#heading=h.u3vzh3jcb8mu) is:
> We want to make sure that exception handling is identical for all the drivers, so it is easier to reason about exceptions on a cross-driver basis.

There are language-imposed differences that prevent the experience from being identical, but wherever possible, the experience should be the same.

I'm not sure why the `position` field is deemed `experimental`. It has existed for more than 3 years, and is key to debugging queries. If you have a query that fails that uses the same function multiple times, knowing which instance of the function is causing the problem is the only thing that lets you target your investigation.

parkhomenko: I've updated the code in this PR: https://github.com/fauna/faunadb-csharp/pull/176
so it looks identical to JVM now,
as well as I've added `Position` to readme file

vadimLF (approved):
LGTM.

vadimLF (approved):

parkhomenko:
Opened a new one for beta version:
https://github.com/fauna/faunadb-csharp/pull/179

PR #176, opened 2021-11-05
chore(FE-1725): Reworked exceptions code

parkhomenko:
### Notes
Reworked exceptions code so it is identical to the JVM driver's

### Jira ticket
https://faunadb.atlassian.net/browse/FE-1725


vadimLF (approved):
LGTM.

PR #189, opened 2022-02-08
Documentation for 4.2

cleve-fauna:
I believe this should be the docs for 4.2.0.

I generated it by running `doxygen doc/Doxyfile` while on the v4 branch and then copying the output for html here.


faunaee (approved):
I opened the PR's branch locally, and browsed to a number of sections within the `4.2.0` folder, and they look good.

I also ran `htmltest` to verify all of the links. They check out fine.

So, LGTM! 👍

cleve-fauna:
not needed just had to unapuse pipeline 🤦

